PR #135, opened 2023-09-19
fix: text wrapping and sticky

ceriverso:


jshearer (approved):
Pushed up one more fix to stop jumpy scrolling, TYVM!

PR #149, opened 2023-10-31
Add redirects from old blogs

SLJEstuary:


jshearer (approved):
Tested, these work. Thanks @SLJEstuary!

PR #150, opened 2023-11-01
Added more redirects

SLJEstuary:


jshearer (approved):
I tested a couple, assuming the rest work 👍

PR #157, opened 2023-11-16
Feat: Blog Posts new 3 column template

flavioriper:
### **Description**
We need this to change the Blog Posts template to 3 columns, including the new Related Posts layout and call to action buttons


src/components/BlogPopularArticles.tsx
```diff
@@ -44,16 +44,14 @@ export const PopularArticles = () => {
...
-                            <li key={index}>
-                                <GatsbyImage
-                                    image={article?.hero?.localFile?.childImageSharp?.gatsbyImageData}
-                                    alt="debezium alternatives"
-                                    className="icon-image popular-articles-image"
-                                />
-                                <Link to={`/${article?.slug}`}>{article?.title}</Link>
-                            </li>
-                        </>
+                        <li key={index}>
```
travjenkins, line 47: This should be switched to something like `populate-articles__${article.slug}`.

Then if the article is actually missing we should just return `null` as there is no reason to include the element if there is not an article.

src/components/BlogPostToc/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+    useEffect(() => {
+        window.addEventListener('scroll', handleScroll, { passive: true });
+        return () => window.removeEventListener('scroll', handleScroll)
+    }, [])
+
+    return (
+        <div className="table-of-contents">
+            <ul>
+                {items.map((item, index) => (
+                    <li className={`${!item.root ? "sub-item" : ""} ${(index === 0 || item.position <= position) ? "active" : ""}`} key={index}>
```
travjenkins, line 25: `key` needs expanded

src/components/CascadingMenu.tsx
```diff
@@ -214,8 +214,8 @@ export const NavMenuList = ({ item }: { item: NavItem }) => {
                     unmountOnExit
                 >
                     <List component="div" disablePadding>
-                        {item.children.map(child => (
-                            <NavMenuList item={child} />
+                        {item.children.map((child, index) => (
+                            <NavMenuList item={child} key={index} />
```
travjenkins, line 218: `key` needs expanded

src/templates/blog-post/Wrapper/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+                            <GatsbyImage
+                                alt={post.title}
+                                image={
+                                    post.hero.localFile.childImageSharp
+                                        .gatsbyImageData
+                                }
+                                loading="eager"
+                            />
+                        ) : null}
+                    </div>
```
travjenkins, line 34: Is there a reason we are still including the `div` as empty when there is no `hero` image?

travjenkins (changes requested):
This branch needs updated. I reviewed it in isolation - so a few of the callouts I think would get fixed by is getting updated.

PR #162, opened 2023-11-24
Feat: Navbar CleanUp and Homepage Tagline Size Fix

flavioriper:
### **Description**
We need this to do some navbar cleanups and adjust homepage tagline size


jshearer (approved):
LGTM, though the navbar does still look squished at certain screen widths
Screen Shot 2023-11-30 at 10 00 31

PR #168, opened 2023-12-01
Feat Case Study Template

flavioriper:
https://www.figma.com/file/yzY87SJ2mEwl1Z7Omt5MBZ/CaseStudy_11_14?node-id=2003%3A79&mode=dev


gatsby-config.ts
```diff
@@ -27,6 +27,23 @@ const strapiConfig = {
         "job-posting",
         "product-comparison-page",
         "employee",
+        {
```
jshearer, line 30: Fancy!

src/layouts/CaseStudy/Head/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+    data: {
+        caseStudy,
+        site: {
+            siteMetadata: { siteUrl },
+        },
+    },
+}) => {
+    return (
+        <>
+            <Seo
```
jshearer, line 17: Excellent

PR #178, opened 2023-12-12
feat: add schema metadata to all pages

flavioriper:


jshearer (changes requested):
Schema.org only understands specific `@type`s, of which `Organization` is one. `Destinations`, `Homepage` etc aren't. See the validation failure: https://validator.schema.org/#url=https%3A%2F%2Festuary-marketing--pr178-feat-schema-metadata-o4sbs50z.web.app%2F

Also, the data for each `@type` is different, so we don't want to combine them like you're doing here. For example, it doesn't make sense for a `BlogPosting` to have an address. See the validation failure here: https://validator.schema.org/#url=https%3A%2F%2Festuary-marketing--pr178-feat-schema-metadata-o4sbs50z.web.app%2Fbigtable-vs-firestore%2F

So what I would do is have `<Seo>` take a `metadata` field that defaults to the `"@type": "Organization"` that you have here. That way e.g `blog-post.tsx` can specify its own metadata, but any page that leaves it unspecified gets the default.

**Note:** Maybe hold off on actually making that change until we hear back from our SEO folks -- it's possible they're going to say that _every_ page should have the `"@type": "Organization"`, in which case the logic would change to optionally including any provided `metadata`, in addition to the Organization one.

jshearer:
Ok so we heard back from SEO folks: _every single page_ should have the Organization type, in addition to any pages that have their own types like blog posts.

jshearer:
I just took another look. To clarify, we want:
* Every single page to have a `<script type="application/ld+json">{"@type": "Organization",…}</script>` with the fields I mentioned in the issue
* The pages that already have existing metadata tags to still have those tags. That is, these pages will have _multiple_ metadata tags:
  * [`blog-post.tsx`](https://github.com/estuary/marketing-site/blob/ce0492de8ebf1e6820215a357aad570ea5190267/src/templates/blog-post.tsx#L160-L188)
  * [Case Studies](https://github.com/estuary/marketing-site/blob/ce0492de8ebf1e6820215a357aad570ea5190267/src/layouts/CaseStudy/Head/index.tsx#L26-L49)

jshearer (approved):
Thanks, that should do it!

PR #179, opened 2023-12-21
Feat: add Osano MGMT

flavioriper:
https://cmp.osano.com/16CPXbTOi1sXx4D3/1e6b223c-ed10-4c4b-a442-48fea69f76af/osano.js


jshearer (commented):
I'm going to hold off on approving this until @dyaffe says go for it, since it also adds Hotjar again. We've added and removed hotjar in the past because it hurts our google pagespeed numbers when we add it, so I don't want to go back and forth again. If you throw up a PR that _just_ adds Osano I'll approve it now 👍

dyaffe:
Why are we adding Hotjar?  As Joseph mentioned, I'm pretty anxious about adding that in due to page speed issues with Google SEO metrics.

flavioriper:
Removed the hotjar until we figure out a better solution to it

gatsby-config.ts
```diff
@@ -125,6 +125,12 @@ const cfg: GatsbyConfig = {
                 },
             },
         },
+        {
```
jshearer, line 128: Ok I heard back from @dyaffe: can we remove this `vwo` stuff for now please? Once that's done this is good to go

jshearer (changes requested):

gatsby-config.ts
flavioriper: vwo removed

gatsby-config.ts
```diff
@@ -125,6 +125,12 @@ const cfg: GatsbyConfig = {
...
+            resolve: "gatsby-plugin-vwo-smartcode",
```
jshearer, line 129: This still needs to be removed, build will fail otherwise

jshearer (changes requested):

jshearer (approved):
LGTM

PR #194, opened 2024-02-05
feat: Add webinar reminder to the blue top bar

flavioriper:
### Description
We need this to change some content talking about Webinar


jshearer (approved):
LGTM

PR #216, opened 2024-03-11
Fix: GA cors error

flavioriper:
## Changes

-   Changed the GA script to be pulled on each build to being hosted locally. This will fix the GA cors issue.

## Tests / Screenshots

image

## Lighthouse

image


static/gtag.js
travjenkins, line 1: We should not just be inline-ing this script like this. If we want to solve this problem this way we need to fetch the tag during build so it stays up to date.

Also - we need to get an idea of how often this changes as this could end up meaning we use an old version of this file and cause other issues.

gatsby-config.ts
```diff
@@ -104,14 +104,13 @@ const cfg: GatsbyConfig = {
     flags: {
         PARALLEL_SOURCING: true,
     },
+    partytownProxiedURLs: [`https://www.googletagmanager.com/gtag/js?id=G-P1PZPE4HHZ`],
```
travjenkins, line 107: We are we setting up a proxied url here? 

Also - we should make `GA_MEASUREMENT_ID` be reusable everywhere in the app so we are not hardcoding it in one spot and not in others. If that is too tough then it needs commented and called out this is hardcoded and needs to stay in sync.

flavioriper: I will remove this proxy, I was trying to find a solution without hosting the script ourselves.

gatsby-node.ts
```diff
@@ -414,5 +415,11 @@ export const createResolvers: GatsbyNode["createResolvers"] = async ({
...
 
 export const onPreBuild = async () => {
-  await copyLibFiles(path.join(__dirname, 'static', '~partytown'));
+    await copyLibFiles(path.join(__dirname, 'static', '~partytown'));
+};
+
+export const onPreBootstrap = async () => {
+    const response = await fetch('https://google-analytics.com/analytics.js');
+    const buffer = Buffer.from(await response.arrayBuffer());
+    await writeFile(path.join(__dirname, 'static', 'analytics.js'), buffer);
```
travjenkins, line 424: This needs very basic error handling.

As a fast follow we need to include a `sha` to compare against so we know when this changes / we got the right file.

flavioriper:
> One issue with Partytown and GA is this break the testing tool : [BuilderIO/partytown#72](https://github.com/BuilderIO/partytown/issues/72)
> 
> We should probably look into that a bit to see if there are any solutions/ideas/work arounds in that issue

Since our GA script is completely down we can't check if it's happening to us, so I would suggest moving on to fix the problem we have now.

what do you think?

gatsby-node.ts
```diff
@@ -414,5 +415,11 @@ export const createResolvers: GatsbyNode["createResolvers"] = async ({
 }
 
 export const onPreBuild = async () => {
-  await copyLibFiles(path.join(__dirname, 'static', '~partytown'));
+    await copyLibFiles(path.join(__dirname, 'static', '~partytown'));
+};
+
+export const onPreBootstrap = async () => {
+    const response = await fetch('https://google-analytics.com/analytics.js');
```
travjenkins, line 422: Current: `https://www.googletagmanager.com/gtag/js?id=G-P1PZPE4HHZ`
New: `https://google-analytics.com/analytics.js`

These two files are not the same thing. What is driving us to no longer load tag manager? That can include more than just analytics.

travjenkins (changes requested):
Some call outs and required changes

static/gtag.js
flavioriper: Since the GA script is not something that changes that much (it barely changes tbh), I say we could inline it for now, at least to solve our priority which is the GA missing metrics. 

To change for something that fetches the tag during build would require a lot of effort and tests

flavioriper:
> > > One issue with Partytown and GA is this break the testing tool : [BuilderIO/partytown#72](https://github.com/BuilderIO/partytown/issues/72)
> > > We should probably look into that a bit to see if there are any solutions/ideas/work arounds in that issue
> > 
> > 
> > Since our GA script is completely down we can't check if it's happening to us, so I would suggest moving on to fix the problem we have now.
> > what do you think?
> 
> This tool is to be used locally.... I do not think this is impacted by the prod issue.

image

it's connecting here, so we probably fine

flavioriper:
Moved GA to be hosted statically within the project like before since it's something that barely changes over the year.

travjenkins (approved):
lgtm

We'll need some stuff built out to make this a proper solution as a fast follow.

Do not merge until a full build/test deploy is double checked.

flavioriper:
The build/test deployment is completed and healthy

can we go for the merge? @travjenkins

PR #217, opened 2024-03-13
feat: Change firebase special character rule

flavioriper:
## Changes

-   Updated the URI-encoded characters from the redirect rule so they can work properly on firebase.

- https://estuary-marketing--pr217-feat-redirect-rules-j030983w.web.app/data-lake-architecture%20/ -> https://estuary-marketing--pr217-feat-redirect-rules-j030983w.web.app/data-lake-architecture
- https://estuary-marketing--pr217-feat-redirect-rules-j030983w.web.app/data%20ingestion-vs-data-integration/ -> https://estuary-marketing--pr217-feat-redirect-rules-j030983w.web.app/data-ingestion-vs-data-integration

image
image

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -97,12 +97,12 @@
                 "type": 301
             },
             {
-                "source": "/data-lake-architecture%20/",
+                "source": "/data-lake-architecture /",
                 "destination": "/data-lake-architecture",
                 "type": 301
             },
             {
-                "source": "/data%20ingestion-vs-data-integration/",
+                "regex": "/data%20ingestion-vs-data-integration/",
                 "destination": "/data-ingestion-vs-data-integration",
                 "type": 301
             }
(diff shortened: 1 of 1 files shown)


firebase.json
```diff
@@ -92,17 +92,17 @@
                 "type": 301
             },
             {
-                "source": "/data%20ingestion-vs-etl/",
+                "source": "/data ingestion-vs-etl/",
```
travjenkins, line 95: I am assuming there are paths used to fetch something... is there a reason we have spaces in them?

flavioriper: @SLJEstuary can probably answer that one better than me, but since it's a redirect rule, I would say this is probably an endpoint that is being accessed for some reason (wrong campaigns, wrong links, etc), and Estuary is making sure even those wrong paths will lead to the right place

travjenkins (approved):
lgtm

PR #220, opened 2024-03-14
Fix: Add key to blogIndex maps

flavioriper:
## Changes

-   Added a key prop to every list items within the component

## Tests / Screenshots

image

### DIFF
--- a/src/templates/blog.tsx
+++ b/src/templates/blog.tsx
@@ -107,8 +107,9 @@ const BlogIndex = ({
                 </div>
                 <div className="blogs-index-tab-bar">
                     <div className="blogs-index-tabs">
-                        {tabCategories.map(category => (
+                        {tabCategories.map((category, index) => (
                             <Link
+                                key={`${category.Slug}-${index}`}
                                 to={`/blog/${category.Slug}`}
                                 className={clsx("blogs-index-tab", {
                                     "blogs-index-tab-active":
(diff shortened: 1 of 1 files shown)


src/templates/blog.tsx
```diff
@@ -131,8 +132,8 @@ const BlogIndex = ({
                     </div>
                 </div>
                 <div className="blogs-index-body">
-                    {(query.length > 0 ? results : posts).map(post => (
-                        <BlogPostCard {...post} />
+                    {(query.length > 0 ? results : posts).map((post, index) => (
+                        <BlogPostCard key={index} {...post} />
```
travjenkins, line 136: `index` is not usually a good key. 

`post` has the `slug` property like above. Please use that.

src/templates/blog.tsx
```diff
@@ -107,8 +107,9 @@ const BlogIndex = ({
                 </div>
                 <div className="blogs-index-tab-bar">
                     <div className="blogs-index-tabs">
-                        {tabCategories.map(category => (
+                        {tabCategories.map((category) => (
                             <Link
+                                key={`${category.Slug}`}
```
travjenkins, line 112: the `Slug` is already a string.... I _think_ we're safe just outputting it.

flavioriper: We can't because of the Typescript rules, the key is not expecting a string, but a React.key typo. We would have to change the way the category is being interacted with to make it work without this

travjenkins: It is expecting `string` type and not `String`. So update the type to `string` and it will be fine.

flavioriper: I did change just the type of slug prop to avoid breaking something else

PR #229, opened 2024-03-18
Feat: Eslint and prettier project setup

flavioriper:
## Changes

- Add Eslint config files
- Add Prettier config files
- Changed `tsconfig `file to sanitize unused rules

## Tests / Screenshots

image
image


tsconfig.json
```diff
@@ -1,101 +1,23 @@
...
-      // "noImplicitOverride": true,                       /* Ensure overriding members in derived classes are marked with an override modifier. */
-      // "noPropertyAccessFromIndexSignature": true,       /* Enforces using indexed accessors for keys declared using an indexed type */
-      // "allowUnusedLabels": true,                        /* Disable error reporting for unused labels. */
-      // "allowUnreachableCode": true,                     /* Disable error reporting for unreachable code. */
-      /* Completeness */
-      // "skipDefaultLibCheck": true,                      /* Skip type checking .d.ts files that are included with TypeScript. */
-      "skipLibCheck": true                                 /* Skip type checking all .d.ts files. */
+        "target": "esnext",
+        "lib": ["dom", "dom.iterable", "esnext"],
+        "jsx": "preserve",
```
travjenkins, line 5: Why are we changing this?

tsconfig.json
```diff
@@ -1,101 +1,23 @@
...
+        "allowJs": true,
```
travjenkins, line 6: I don't think this is needed as we don't have any `js` files I can see in our source

tsconfig.json
flavioriper: I am going for something more "clean", so I decided to remove all of the commented variables (there are a lot of documents talking about each of them so it's easy to find them). And I made a few changes following good typescript SSR patterns to not conflict with Eslint or Prettier patterns.

tsconfig.json
flavioriper: It's good to have this rule on since we use self-baked plugins and patches that could use bare js, and would be easy to integrate them without changing the rules in the future.

tsconfig.json
travjenkins: Why are we changing the `jsx` property?

tsconfig.json
travjenkins: If we are using things we wrote then they should never be in `js`.
If we are writing patches then we do not want linting rules applied because we need to follow whatever is in the package to begin with.

.eslintrc.js
```diff
@@ -0,0 +1,32 @@
+/* eslint-env node */
+
+module.exports = {
```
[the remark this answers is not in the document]
flavioriper: I've tried to use the Gatsby plugin to handle it, but none of them worked. So I decided to go for the bare Eslinting flow, this way even if in the future the framework change, those rules would work well too

tsconfig.json
flavioriper: image

the preserve outputs more common files, but I can go for React, I don't think would be a problem either

tsconfig.json
flavioriper: Sure, I will change it to false

tsconfig.json
travjenkins: I know what it does. I am asking why was this changed?

flavioriper: I'm moving it back to react

tsconfig.json
flavioriper: Done

travjenkins (changes requested):
Can you please add this option:
`'react-hooks/exhaustive-deps': 'error'`

Also, we can ignore for now but we should probably add the rule `'react/jsx-no-leaked-render': 'error'` in there. I see we use this approach a lot but it is somewhat risky to be doing.

travjenkins (approved):
lgtm - we'll make a follow up PR where these are fixed

Issue #235, opened 2024-03-20
Add Osano back into marketing site


travjenkins:
## Description

describe what you would like changed

## Work required

-  [x] review point in the [performance docs](https://docs.osano.com/hc/en-us/articles/22472101615764-Performance-Enhancement-and-Size-Reduction-2021) and make sure we're making this efficient
-  [x] add Osano onto the website
-  [x] run Lighthouse tests before merging and seeing what the impact will be
-  [x] review impact and make a decision
     -  [ ] Mark scripts with inline-classifications for Osano

## Figma / Designs

N/A

travjenkins:
For some reason - Gatsby renders page **really** different for develop mode

![image](https://github.com/estuary/marketing-site/assets/270078/7594a916-82e4-44b2-8b86-f6a5e60f2856)

This means that when running dev locally your `onPreRenderHTML` won't **EVER** get access to the list of scripts being loaded. This makes testing out changes here really annoying.

Why they do this - I have no clue.

travjenkins:
The normal static generator will make sure you have access to all the script tags before sending over. I assume they want to make sure a user does not accidentally mess with the files that are required for development... but this seems like a bad trade off as it makes developing so much harder.

![image](https://github.com/estuary/marketing-site/assets/270078/94a2bb74-2c19-4b70-bd2c-6ce289aae8ff)

travjenkins:
Adding a handler to `onPreRenderHTML` might not work :( https://github.com/gatsbyjs/gatsby/discussions/37036

travjenkins:
So adding in the `onPreRenderHTML` function won't work as we do not get the scripts loading back and just the slices.

![image](https://github.com/estuary/marketing-site/assets/270078/ba11577a-d50f-4b22-82e2-b95be5d4496b)

travjenkins:
https://github.com/estuary/marketing-site/issues/250 will replace this for now.

We can switch back to Osano once we can use inline classification (Switching to Next.JS)

PR #236, opened 2024-03-20
Enable Osano again

travjenkins:
## Changes

Fixes https://github.com/estuary/marketing-site/issues/235

- Add back in Osano tag
- Add inline classifications for Osano where we can
- Break up a Section Two as it was called out in lighthouse as slow
- Added Osano to the preconnect.... not even sure if this will help
- Enabled `Split Payload` on [Osano](https://docs.osano.com/hc/en-us/articles/22471437258900-Split-Payload-Setting)

## Tests / Screenshots

- TBD


gatsby-browser.tsx
```diff
@@ -1,89 +1,85 @@
...
+      <>
+        {element}
+        <Script
+          id="ze-snippet"
+          key="gatsby-plugin-zendesk-chat"
+          strategy={'idle'}
+          async
+          defer
+          src={`https://static.zdassets.com/ekr/snippet.js?key=${ZD_KEY}`}
+          data-osano="ESSENTIAL"
```
travjenkins, line 40: labeling for Osano

gatsby-browser.tsx
```diff
@@ -1,89 +1,85 @@
...
+        />
+        <Script
+          id="hs-script-loader"
+          async
+          defer
+          strategy={'idle'}
+          src="//js.hs-scripts.com/8635875.js"
+          data-osano="MARKETING"
```
travjenkins, line 48: labeling for Osano

patches/react-hubspot-form+1.3.7.patch
```diff
@@ -11,3 +11,11 @@ index 7f1ffa5..49c174b 100644
  						_this2.props.onSubmit(formData);
  					}
  				});
+@@ -82,6 +82,7 @@ var HubspotForm = function (_React$Component) {
+ 				_this3.createForm();
+ 				_this3.findFormElement();
+ 			};
++			script['data-osano'] ="ESSENTIAL";
```
travjenkins, line 18: labeling for Osano

src/components/SectionTwo/index.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react';
+import SectionTwoHeader from './SectionTwoHeader';
+import SectionTwoHero from './SectionTwoHero';
+import SectionTwoTiles from './SectionTwoTiles';
+
+const SectionTwo = () => {
```
travjenkins, line 6: Trying to break this up a bit as it was called out as a slow painting.

flavioriper: Do we have some strategy on the code to split the js bundle properly?

travjenkins: Nothing defined yet. Right now I use Lighthouse to see if something loaded slowly. Then I just try to figure out if it can be split up in _any_ way.

PR #242, opened 2024-04-03
Trying to get Google Analytics working better

travjenkins:
## Changes

- Remove Partytown as it is causing CORS issues and the proxied URL seems to not help
- Go back to loading `gtag` from Google

## Tests / Screenshots

- ...


gatsby-config.ts
```diff
@@ -4,132 +4,131 @@
...
+    social: {
+      twitter: 'estuary twitter',
     },
-    flags: {
-        PARALLEL_SOURCING: true,
+  },
+  flags: {
+    PARALLEL_SOURCING: true,
+  },
+  partytownProxiedURLs: [GTAG_PATH],
```
travjenkins, line 108: This is the only true change

gatsby-config.ts
travjenkins, line 1: This is just formatting. Happened because I initially had a change in this file.

.gitignore
travjenkins, line 1: Partytown Cleanup

gatsby-ssr.tsx
```diff
@@ -11,36 +13,19 @@ const { Partytown } = require("@builder.io/partytown/react")
...
-            src={`/gtag.js?id=${GA_MEASUREMENT_ID}`}
-        />,
-        <script
-            key="google-analytics-config"
-            type="text/partytown"
-            dangerouslySetInnerHTML={{
-                __html: `window.dataLayer = window.dataLayer || [];
+  setHtmlAttributes({ lang: `en` });
+  setHeadComponents([
+    <script key="google-analytics" src={GTAG_PATH} />,
```
travjenkins, line 19: We need to load this script directly from Google

package.json
travjenkins, line 1: Partytown Cleanup

static/gtag.js
travjenkins, line 1: Partytown Cleanup (do not want to serve this file ourselves)

gatsby-browser.tsx
```diff
@@ -1,89 +1,53 @@
...
-      requestAnimationFrame(() => {
-        requestAnimationFrame(() => setTimeout(sendPageView, 0))
-      })
-    } else {
-      // Delay by 32ms to simulate 2 requestOnAnimationFrame calls
-      setTimeout(sendPageView, 32)
-    }
-  
-    return null
-  }
```
travjenkins, line 89: This is the main change. Removing this as the plugin handles it.

gatsby-ssr.tsx
```diff
@@ -11,36 +10,6 @@ const { Partytown } = require("@builder.io/partytown/react")
  * @type {import('gatsby').GatsbySSR['onRenderBody']}
  */
 
-const GA_MEASUREMENT_ID = "G-P1PZPE4HHZ"
```
travjenkins, line 14: Moved this to `shared.ts`

gatsby-ssr.tsx
```diff
@@ -11,36 +10,6 @@ const { Partytown } = require("@builder.io/partytown/react")
...
-            type="text/partytown"
-            dangerouslySetInnerHTML={{
-                __html: `window.dataLayer = window.dataLayer || [];
-        window.gtag = function gtag(){ window.dataLayer.push(arguments);}
-        gtag('js', new Date()); 
-        gtag('config', '${GA_MEASUREMENT_ID}', { send_page_view: false })`,
-            }}
-        />,
-        // <script key="osano-1" src="https://cmp.osano.com/16CPXbTOi1sXx4D3/1e6b223c-ed10-4c4b-a442-48fea69f76af/osano.js"></script>,
-    ])
```
travjenkins, line 45: Removed Partytown and switched back to gtag plugin

gatsby-browser.tsx
```diff
@@ -1,89 +1,53 @@
...
-}
+  const [dimensions, setDimensions] = useState({ width: 0, height: 0 });
+  React.useEffect(() => {
+    const subscriber = () => setDimensions({ width: window.innerWidth, height: window.innerHeight });
+    window.addEventListener('load', subscriber);
+    return () => window.removeEventListener('load', subscriber);
+  }, []);
+  if (process.env.NODE_ENV === 'development' || isMobile || dimensions.width < 768) {
+    return element;
+  } else {
```
Brenosalv, line 29: Nothing impactful, but the "else" is not necessary here.

gatsby-ssr.tsx
Brenosalv: @travjenkins In the big-query PR, I created a folder called "constants" in the src folder to store all the shared constants. Under constants folder, I created a file called "urls.ts" to store all the shared urls. It's good to separate the types of shared constants. The "GTAG_PATH" could be stored in the "urls.ts" file and the GA_MEASUREMENT_ID in a "keys.ts" or "ga.ts" in case this ID is public. What do you think about it?

travjenkins: No - I think that is way too much separation that does not really solve a problem we have. If we end up with tons of keys, urls, etc. then breaking them out can make sense to me. Otherwise we just end up with a lot of files that are not super helpful.

gatsby-browser.tsx
travjenkins: I never even looked at the code - good point. I'll clean that up.

gatsby-ssr.tsx
Brenosalv: Okay, I will adopt the shared.ts then.

Brenosalv (dismissed):

gatsby-ssr.tsx
Brenosalv: Yes, I created the 'constants' folder and 'urls.ts' file with scalability in mind, but I agree we do not need that at the moment.

shared.ts
```diff
@@ -0,0 +1,3 @@
+export const GA_MEASUREMENT_ID = 'G-P1PZPE4HHZ';
+
+export const GTAG_PATH = `https://www.googletagmanager.com/gtag/js?id=${GA_MEASUREMENT_ID}`;
```
jshearer, line 3: I may be missing it, but is this used anywhere now that the gatsby plugin handles injecting `gtag.js`? AFAICT it's unused

gatsby-config.ts
```diff
@@ -4,132 +4,148 @@
...
-                ],
-            },
+  },
+  flags: {
+    PARALLEL_SOURCING: true,
+  },
+  // graphqlTypegen: true,
+  plugins: [
+    {
+      resolve: `gatsby-plugin-google-gtag`,
```
jshearer, line 111: Looks like this file got reformatted, but AFAICT this is the only actual change in here, which looks right to me 👍 

One thought I just had: does `gatsby-plugin-google-gtag` preload the `googleanalytics.com` domain, or wherever `gtag.js` comes from? I assume so, but if not it'd be easy to add to `gatsby-plugin-preconnect` down below

jshearer (approved):
This all LGTM

shared.ts
travjenkins: Oh duh - I'll remove this.

PR #248, opened 2024-04-11
Removing the old links and setting back to the demo

travjenkins:
Also preparing for this to open the banner again when things change

## Changes

- Remove Feb 10 webinar link
- Add back in link to `Demo`
- Add support for the banner to open when a new webinar is added

## Tests / Screenshots

- Manually hit the home page. Nothing shows because I had dismissed it and we still use the old key to check if the user has dismissed already
image

- Altered the value to cause the demo to show up
image

- Added a fake webinar
image
image

### DIFF
--- a/src/components/TakeATour/index.tsx
+++ b/src/components/TakeATour/index.tsx
@@ -4,30 +4,32 @@ import CloseIcon from '@mui/icons-material/Close';
 
 import { Button, Container, Link } from './styles';
 
-const STORAGE_NAME = '@estuary/closeTour';
-
-// If there is an upcoming webinar: `webinar_{date}`
-// If there is no webinar: `demo`
-const LATEST_VERSION = 'demo';
-
-// defaults
-// href: '/why',
-// message: 'Take a Product Tour',
-const SETTINGS = {
+const DEFAULTS = {
   href: '/why',
   message: 'Take a Product Tour',
+  version: '',
 };
+
+const SETTINGS = {
+  ...DEFAULTS,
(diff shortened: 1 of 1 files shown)


src/components/TakeATour/index.tsx
```diff
@@ -1,29 +1,41 @@
...
+const SETTINGS = {
+  ...DEFAULTS,
+  // When there is a webinar uncomment below and provide the necessary details
+  // href: 'fake/path/name',
+  // message: 'This is a fake future Webinar',
+  // version: `/webinar_YYYY/MM/DD`,
+};
 
-const STORAGE_NAME = '@estuary/closeTour';
+const STORAGE_KEY = `@estuary/closeTour${SETTINGS.version}`;
```
travjenkins, line 19: This is the quickest solution - but does potentially cause issues years from now where there is a TON of old webinars in local storage. We might need to add some basic code that cleans up the _previous_ setting if it is there - that way it never gets too full.

We _could_ work these settings into local storage itself but that is a lot of work and would make upgrading harder in the future.

Brenosalv: What do you think about setting an expiration date to each stored value and always or periodically check this date. If it's expired (current date greater than expiration date), it's automatically removed from the local storage.

travjenkins: I had a similar thought - but then we're kind of just reimplementing cookies manually in a way. 

I'll think about it some more.

Issue #254, opened 2024-04-17
Add the webinar pop up to the new blog page


travjenkins:
## Description

The blog template design contained a new webinar sign up pop over. 

## Work required

- [x] Decide if this will truly only be for the blog page template
- [x] Decide if we will show the 101 link, the sign up form, or some combination...
    - make pop over "smart" and show the sign up form if upcoming one is coming
    - only show 101 link if there are no upcoming webinars
    - based on cookies/local storage where we only show if a user has not signed up for one
- [x] TDB - based on the decision above

altitude-lucas:
@Brenosalv a few notes on the decisions needed above:

1. Let's show this popup on the blog page, and any blog post pages
2. Let's make this a smart popup - if there is an upcoming webinar, we show a form to register, if there is not, we show a link to watch the most recent video

altitude-lucas:
@Brenosalv - Chatted with the team, we're going to make an adjustment here, rather than using the smart popup, we're going to use design 1537, and have it link to the static page for the "Change Data Capture 101" page and webinar. 

This will act as a measurement tool for the value the popup provides, and in time we may develop a live/smart form that updates as we release details about the upcoming webinars.

Brenosalv:
@altitude-lucas Thank you for the clarification.

PR #269, opened 2024-04-26
Brenosalv/feature/add webinar popup to the blog page

Brenosalv:
## Issues

- [254](https://github.com/estuary/marketing-site/issues/254)

## Changes

-   Add blog page popup

## Figma layout

- https://www.figma.com/file/qhXFvOJPqLgGsr9lsAPYbn/Estuary%3A-Product-Page?type=design&node-id=3291-5881&mode=design&t=9nfMhmDPM6JWd9Nr-0

## Tests / Screenshots

image


Brenosalv:
For now we decided to implement only this popup version:
image

src/components/BlogPostPopupModal/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+import * as React from "react"
+import { webinarsUrl } from "../../../shared"
+import EstuaryLogo from "../../svgs/colored-logo.svg"
+import { CloseButtonWrapper, Container, LeftColumn, NoThanksButton, RightColumn, Subtitle, Title, WatchNowLink, XIcon } from "./styles"
+
+function BlogPostPopupModal() {
+  const [openDialog, setOpenDialog] = React.useState(false);
+
+  React.useEffect(() => {
+    let hasOpened = false;
```
travjenkins, line 14: Instead of a `let` in an effect I think we should store this within a `ref` to be safer.

src/components/BlogPostPopupModal/styles.ts
travjenkins, line 1: When on a narrow and short view the modal gets a little messed up. This is not a massive deal - but we should try to get the parent to scroll OR the modal not go under the header navigation.

image

Brenosalv: Yeah, I forgot to test it in a real device height. I'll fix this.

src/components/BlogPostPopupModal/styles.ts
```diff
@@ -0,0 +1,141 @@
...
+  margin-bottom: -54px;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  gap: 16px;
+  text-align: center;
+
+  & > span:nth-child(2) {
+    font-size: 1.25rem;
+  }
```
travjenkins, line 73: We should not be targeting elements like this as it makes the styling very brittle and easy to break.

src/components/BlogPostPopupModal/styles.ts
```diff
@@ -0,0 +1,141 @@
...
+
+export const WatchNowLink = styled.a`
+  color: #ffffff;
+  background-color: #5072eb;
+  padding: 16px 24px;
+  width: 100%;
+  border-radius: 4px;
+  margin-bottom: 16px;
+  text-align: center;
+  cursor: pointer;
```
travjenkins, line 110: Since this is a styled anchor tag I do not think we need to `cursor' setting.

src/components/BlogPostPopupModal/styles.ts
```diff
@@ -0,0 +1,141 @@
...
+`
+
+export const NoThanksButton = styled.button`
+  color: #47506d;
+  background-color: transparent;
+  padding: 16px 24px;
+  width: 100%;
+  border: 2px solid #5072eb;
+  border-radius: 4px;
+  cursor: pointer;
```
travjenkins, line 120: Same comment about `cursor` not being needed.

src/components/BlogPostPopupModal/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+        setOpenDialog(true);
+        hasOpened = true;
+      }
+    };
+
+    window.addEventListener('mouseout', handleExitIntent);
+
+    return () => {
+      window.removeEventListener('mouseout', handleExitIntent);
+    };
```
travjenkins, line 27: Just making sure - this seems to only fire when a user moves outside of the window within the browser. If I move between the browser window and my IDE (while both being open) nothing happens. I think this is best because it let's people reference articles more easily - just wanted to make sure that as intentional.

src/components/BlogPostPopupModal/styles.ts
```diff
@@ -0,0 +1,141 @@
...
+    padding: 24px;
+  }
+`
+
+export const Title = styled.span`
+  font-size: 2rem;
+  font-weight: 700;
+  text-transform: uppercase;
+  margin: 50px 96px 0 96px;
+  margin-bottom: -54px;
```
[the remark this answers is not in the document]
Brenosalv: It was a mistake. I'll review my CSS code harder next time. About the negative margin bottom, it's to compensate the margin top of the image exported from Figma. The image only comes with this margin top by default.

src/components/BlogPostPopupModal/styles.ts
Brenosalv: I could use the ```OutboundLink``` component instead of a button here. What do you think?

src/components/BlogPostPopupModal/styles.ts
Brenosalv: iPhone 11:
image

iPhone 5/SE:
image

iPhone 14 pro max:

src/components/BlogPostPopupModal/styles.ts
Brenosalv: Consider the same answers as above.

Brenosalv: Without the cursor pointer, the "No, thanks" button will behave like a disabled button not indicating it's clickable.

src/components/BlogPostPopupModal/index.tsx
Brenosalv: Yes, that's intentional. The only reason to open the popup is when the user tries to leave the blog post, which happens when they try to switch to another tab or close the tab/window, and it's all related to moving the cursor to the browser header/navbar.

src/components/BlogPostPopupModal/styles.ts
travjenkins: That works - that is also probably better because it is linking to another domain.

travjenkins (commented):
lgtm - still need to test in preview before approving.

travjenkins (approved):
lgtm

PR #271, opened 2024-04-29
Adding in CookieFirst integrations

travjenkins:
## Changes

- Add CookieFirst script

## Tests / Screenshots

- N/A

### DIFF
--- a/gatsby-ssr.tsx
+++ b/gatsby-ssr.tsx
@@ -1,3 +1,5 @@
+import { GA_MEASUREMENT_ID } from './shared';
+
 const React = require('react');
 
 /**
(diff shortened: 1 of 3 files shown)


gatsby-ssr.tsx
```diff
@@ -10,6 +11,46 @@ const React = require('react');
...
+      // Default consent settings and tell it to wait a little bit for an wait
+      //  for an update that will be coming from the banner module
+      gtag('consent', 'default', {
+          'ad_storage': 'denied',
+          'ad_user_data': 'denied',
+          'ad_personalization': 'denied',
+          'analytics_storage': 'denied',
+          'functionality_storage': 'denied',
+          'personalization_storage': 'denied',
+          'security_storage': 'denied',
```
Brenosalv, line 29: Before the ```security_storage``` value was set to true. Should it be set to false now?

Brenosalv (approved):
Lgtm. Just left a comment.

gatsby-ssr.tsx
travjenkins: To be safe - I think so. Some articles show it as being forced on. However, I have seen others mark it as truly optional. For our stuff it does not really matter so I think we're okay.

Issue #265, opened 2024-04-24
Breadcrumbs - Connectors


altitude-lucas:
In an effort to ease site usability, I'm recommending that we add Breadcrumbs to the Connector pages, between the H1 + H2

Brenosalv:
@altitude-lucas It needs design work yet, right? If so, we can add the ```needs design``` label to this issue.

altitude-lucas:
![Image](https://github.com/estuary/marketing-site/assets/167227013/207a9ae5-d2e4-4def-a3c4-9c3ab126ff62)

@Brenosalv - from Design

altitude-lucas:
@Brenosalv should be good to push here as well

Issue #264, opened 2024-04-24
Breadcrumbs - Blog Pages


altitude-lucas:
In an effort to ease site usability, I'm recommending that we add Breadcrumbs to the Blog pages, between the H1 + H2

altitude-lucas:
![Image](https://github.com/estuary/marketing-site/assets/167227013/94cbbb3c-17b2-484e-afab-380fc03dd4e4)

@Brenosalv - From Design

altitude-lucas:
@Brenosalv - I looked at the test site here, should be good to go on my end

Brenosalv:
@altitude-lucas Yeah, just waiting Travis to approve the merge.

PR #283, opened 2024-05-03
Brenosalv/feature/breadcrumbs-blog-pages

Brenosalv:
Solves issues #264 and #265

### DIFF
--- a/src/components/Breadcrumbs/index.tsx
+++ b/src/components/Breadcrumbs/index.tsx
@@ -1,0 +1,35 @@
+import * as React from 'react';
+import { Container, CurrentPageTitle, Divider, Links, PreviousLink } from './styles';
+
+interface BreadcrumbsProps {
+  breadcrumbs: {
+    id: string
+    title: string
+    href?: string
+  }[]
+}
+
+const Breadcrumbs = ({ breadcrumbs }: BreadcrumbsProps) => {
+  return (
+    <Container aria-label="breadcrumb">
+      <Links>
+        {breadcrumbs?.map((breadcrumb) => (
+          <li key={breadcrumb.id}>
+            {breadcrumb?.href ? (
+              <>
+                <PreviousLink to={breadcrumb.href}>
+                  {breadcrumb.title}
+                </PreviousLink>
+                <Divider> / </Divider>
+              </>
+            ) : (
+              <CurrentPageTitle>{breadcrumb.title}</CurrentPageTitle>
+            )}
+          </li>
+        ))}
+      </Links>
+    </Container>
+  )
+}
+
+export default Breadcrumbs
--- a/src/components/Breadcrumbs/index.tsx
+++ b/src/components/Breadcrumbs/index.tsx
@@ -15,7 +15,7 @@ const Breadcrumbs: React.FC<BreadcrumbsProps> = ({ breadcrumbs = [] }) => {
     <Container aria-label="breadcrumb">
       <Links>
         {breadcrumbs.map(({ title, href }) => (
-          <Crumb key={href || title}>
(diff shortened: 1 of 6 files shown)


src/components/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+import { Container, CurrentPageTitle, Divider, Links, PreviousLink } from './styles';
+
+interface BreadcrumbsProps {
+  breadcrumbs: {
+    id: string
```
travjenkins, line 6: Instead of passing `id` we should just generate the key with the stuff passed in. 

Something like 'breadcrumb_${href}' maybe

src/components/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const Breadcrumbs = ({ breadcrumbs }: BreadcrumbsProps) => {
+  return (
+    <Container aria-label="breadcrumb">
+      <Links>
+        {breadcrumbs?.map((breadcrumb) => (
+          <li key={breadcrumb.id}>
+            {breadcrumb?.href ? (
+              <>
+                <PreviousLink to={breadcrumb.href}>
```
travjenkins, line 20: We should add some styling to make sure if the title is suuuuuuper long things do not wrap or look bad.

travjenkins (commented):
Some small changes

src/components/Breadcrumbs/index.tsx
Brenosalv: Yeah, I kept the breadcrumbs only for screen width greater than 768px. Additionally I will limit the number of characters adding ellipsis for super long titles.

src/components/Breadcrumbs/index.tsx
Brenosalv: I thought about that but the last crumb does not have an href since it's not a link as the others. Even so I can customize to accept this.

src/components/Breadcrumbs/index.tsx
```diff
@@ -1,31 +1,30 @@
...
-const Breadcrumbs = ({ breadcrumbs }: BreadcrumbsProps) => {
+const Breadcrumbs: React.FC<BreadcrumbsProps> = ({ breadcrumbs = [] }) => {
   return (
     <Container aria-label="breadcrumb">
       <Links>
-        {breadcrumbs?.map((breadcrumb) => (
-          <li key={breadcrumb.id}>
-            {breadcrumb?.href ? (
+        {breadcrumbs.map(({ title, href }) => (
+          <Crumb key={href || title}>
```
travjenkins, line 18: Can you make the key something with the component name in it to help reduce the chance of a possible clash.

'breadCrumbs_...'

Brenosalv: Sure, will do.

src/components/Breadcrumbs/styles.ts
travjenkins, line 1: Can you style the breadcrumb to not wrap?

image

travjenkins: Never mind on this. Not that bad and also _could_ make certain widths work better.

travjenkins (approved):
lgtm - we _might_ want to look into a breadcrumb generator eventually to prevent this from getting out of sync or something. However, that is kind of overkill for right now.

src/components/Breadcrumbs/styles.ts
Brenosalv: Yeah, for tablets and other mobile devices it would be necessary for the last crumb to be visible enough.

Issue #275, opened 2024-05-02
Icons disappearing in the real-time section from the connector page


Brenosalv:
**Bug**
Once we reduce the screen width, the icons disappear.

**Expected**
Fix responsiveness for the cards and their icons.

**Screenshots**
![image](https://github.com/estuary/marketing-site/assets/60396753/d4389be0-2eb9-43c7-bcc5-ded38ba8b7b3)

travjenkins:
Was this fixed already? I am not seeing it on the current width preview build
![image](https://github.com/estuary/marketing-site/assets/270078/bc44e775-eae6-41cf-a735-b2bc541b7720)

Brenosalv:
@travjenkins I fixed this along with the max-width. You can check it in the last build here #284 .

Issue #255, opened 2024-04-17
Implement a max width for the site


travjenkins:
## Description

The site can get as wide as any monitor will allow. This means the site can get **really** wide and start looking weird. 

Content within the site (header, body, etc) should be a max of 1280px, and padding on either side should expand as the page gets wider to accommodate monitors of all widths.

Also, the additional padding left on the blog post content should be removed for mobile devices.

## Work required

- [x] agree on max width of 1280
- [x] implement max width and review pages to check where we need to ensure things get padding
- [x] remove the additional padding left on the blog post content for mobile devices.

## Figma / Designs

TBD

altitude-lucas:
@Brenosalv - as part of the acceptance criteria here, can you attach a list of the different page templates that will need to be observed during the review phase?

Brenosalv:
> @Brenosalv - as part of the acceptance criteria here, can you attach a list of the different page templates that will need to be observed during the review phase?

Sure, I'll do that

travjenkins:
I do not want to explode scope on this but this is probably a good time to at least start _documenting_ (if not fixing) the weird responsiveness on some of the pages.

Here is an example from the connectors page: 
![image](https://github.com/estuary/marketing-site/assets/270078/e82c7e23-2771-491d-ac01-bd9572d01240)

Brenosalv:
@travjenkins Sure, I can check the responsiveness on all the pages. If it's quick to solve, I'll fix it; if not, I'll document new issue.

Brenosalv:
@altitude-lucas
Here is the list of pages for review:
- Homepage - https://estuary-marketing--pr284-brenosalv-feature-im-qjveud3g.web.app
- Connectors - https://estuary-marketing--pr284-brenosalv-feature-im-qjveud3g.web.app/integrations
- Blog grid - https://estuary-marketing--pr284-brenosalv-feature-im-qjveud3g.web.app/blog/data-engineering
- About - https://estuary-marketing--pr284-brenosalv-feature-im-qjveud3g.web.app/about
- Podcasts - https://estuary-marketing--pr284-brenosalv-feature-im-qjveud3g.web.app/podcasts
- Product - https://estuary-marketing--pr284-brenosalv-feature-im-qjveud3g.web.app/product
- Case Studies - https://estuary-marketing--pr284-brenosalv-feature-im-qjveud3g.web.app/customers/casestudy/trueplatform
[...]

PR #284, opened 2024-05-04
Brenosalv/feature/implement-a-max-width-for-the-site

Brenosalv:
### Issues:
https://github.com/estuary/marketing-site/issues/255
https://github.com/estuary/marketing-site/issues/275


Brenosalv:
Just correcting: The commit [7dd6061](https://github.com/estuary/marketing-site/pull/284/commits/7dd606150b73ea03c353343fca0e2cdb559f8a38) was for the connector template, not blog post template.

travjenkins (changes requested):
need a few small fixes

src/pages/404.tsx
```diff
@@ -1,19 +1,28 @@
...
+import Layout from "../components/layout";
+import Seo from "../components/seo";
+import { Container } from "./styles";
+
+import styled from 'styled-components';
+import { globalMaxWidth } from '../globalStyles';
+
+const Container = styled.div`
+  ${globalMaxWidth}
+`;
```
travjenkins, line 12: You have two `Container`s

image

Brenosalv: Fixed.

travjenkins (dismissed):
lgtm - would like to wait for another preview build to complete for one last quick look before merging.

Brenosalv:
> If we turn off `nowrap` on the text I think it looks okay - image

Thanks, Travis, that's exactly the solution.

Brenosalv:
> Sorry - found one more while testing on my home machine. image
> 
> We should mark these connector icons with hardcoded width and height? However, that can cause the chips to overflow image
> 
> If we allow the flex items to `wrap` and align them to the `end` it looks okay image
> 
> I am not super strong about any specific solution though - just wanted to leave a basic idea of a quick solution

I see it's an issue already present in production. I like your solution but I prefer besides adding a max-width 55px to the icons, also adjusting the grid to become 3 columns earlier. Please let me know what you think about this solution. I just pushed a new commit.

travjenkins (approved):
lgtm

PR #294, opened 2024-05-14
Adding back in the social links

travjenkins:
## Changes

- Put social links back into header

### Local
image

Like 1 pixel before the area is hidden. Think it looks good as the button is _just_ out of frame and then it hides the normal navigation
image
image

### DIFF
--- a/src/style.less
+++ b/src/style.less
@@ -1,68 +1,66 @@
 /* CSS Custom Properties Definitions */
 
 :root {
-    --maxWidth-none: "none";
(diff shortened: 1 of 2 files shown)


Brenosalv (approved):

src/style.less
travjenkins, line 1: Lint formatted this so here are the true changes

Reduced padding
image

Made social hide with others
image

Brenosalv (approved):

Issue #188, opened 2024-01-26
Redesigned Homepage + New Copy


flavioriper:
## Description
We decided to create a more professional-looking homepage. Our goal is to get higher conversion/engagement time metrics

## Work required

- [x] Updated Copy
- [x] Updated Design

## Figma / Designs
https://www.figma.com/file/qhXFvOJPqLgGsr9lsAPYbn/Estuary%3A-Product-Page?type=design&node-id=1402%3A4330&mode=design&t=azHJzVJR5dJa7A0Q-1

altitude-lucas:
Sent to design for re-vamp with updated copy from the team

altitude-lucas:
@Brenosalv the design is ready to go: https://www.figma.com/file/qhXFvOJPqLgGsr9lsAPYbn/Estuary%3A-Product-Page?type=design&node-id=3627%3A6071&mode=design&t=L0Racx1Yw6MYAiLQ-1

Please make this high-priority for this week

PR #295, opened 2024-05-15
Brenosalv/feature/new homepage features

Brenosalv:
#188 

# Changes

- Implement new homepage according to the figma layout: https://www.figma.com/design/qhXFvOJPqLgGsr9lsAPYbn/Estuary%3A-Product-Page?node-id=0-1&t=X7bBCXEpmg37j3VH-0
- Add different button and link types with hover effect


Brenosalv:
@travjenkins I'm planning to review all the code before you begin your review. However, if you want to start ahead of me, that's fine too. I created the PR now to ensure we have an initial build version available while I review the code. Also, I'd like to address the ESLint issue soon as they're complicating the review process sometimes decreasing our productivity. Ah, and I'm sorry for the large PR.

Brenosalv:
@travjenkins Also, just FYI - I created the branch pointing to the max-width branch to start from there. So feel free to skip the commits related to the max-width. If you review by commit, you can start from the commit [6676c19](https://github.com/estuary/marketing-site/pull/295/commits/6676c198aa17493621dd2fa8404c093f29bbb1e1).

global.d.ts
```diff
@@ -0,0 +1,7 @@
+export { };
```
travjenkins, line 1: There should not be a space in the `{ }`

global.d.ts
```diff
@@ -0,0 +1,7 @@
...
+
+declare global {
+  interface Window {
+    gtag: (command: 'config' | 'event' | 'set' | 'js', targetId: string | Date, config?: any) => void;
```
travjenkins, line 5: This `any` should be `unknown`

src/components/OutboundLink/index.tsx
travjenkins, line 1: I don't think this solution is the way we should go. I think we should just make several components that are all able to be imported instead of passing a variant.

src/components/PricingCalculator/PricingCalculatorSelector.tsx
```diff
@@ -0,0 +1,47 @@
...
+    setSelectedConnectors(value >= 2 ? value : 2)
+  }
+
+  return (
+    <ConnectorsCounterWrapper>
+      <ButtonMinus onClick={handleMinusClick}>
+        <MinusSign />
+      </ButtonMinus>
+      <Form variant="outlined">
+        <InputLabel htmlFor="pricing-calculator-selector-input">Number of connectors</InputLabel>
```
travjenkins, line 28: This label text should be darker... I think it is just straight black in the updates from Polina.
image

src/components/OutboundLink/index.tsx
Brenosalv: Okay, I like this other solution too. I can do it. I was trying to follow how some libraries handle it.

src/components/PricingCalculator/PricingCalculatorSelector.tsx
Brenosalv: Thanks for the heads-up! I hadn't noticed the update.

src/components/Homepage/SectionNine/index.tsx
travjenkins, line 1: We don't need to change in this PR yet - but we should look into getting the carousel switched over to HTML/CSS one.

https://css-tricks.com/css-only-carousel/

src/components/Homepage/SectionNine/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+            return (
+              <Slide key={id} className="keen-slider__slide" style={{ display: !loaded && "none" }}>
+                {isImageSvg ? (
+                  <AvatarSvg src={logo.localFile.publicURL} alt={name + " avatar"} width={110} height={110} />
+                ) : (
+                  <AvatarImg
+                    image={logo.localFile.childImageSharp?.gatsbyImageData}
+                    alt={name + " avatar"}
+                  />
```
travjenkins, line 74: Can you switch the `alt` text to

```
`${name} avatar`
```

src/components/Homepage/SectionNine/index.tsx
Brenosalv: Okay, will leave it for another PR.

Brenosalv: @altitude-lucas Could you create a ticket in the board for this?

Brenosalv:
> We should probably put _some_ minimum width on the page.
> 
> image

I was considering the min screen width a device can have is 320px, so I didn't care about widths less than this. But I can add a ```min-width: 320px```.

travjenkins (dismissed):
LGTM - waiting for one last preview build and final testing before merging

travjenkins (dismissed):
lgtm - still waiting on preview build to merge

travjenkins (approved):
lgtm - ready to merge

PR #307, opened 2024-05-21
Brenosalv/bug/help-chat-not-showing-up-sometimes

Brenosalv:
#308 

## Changes

-   Help Chat is not showing up sometimes when reloading the pages.
-   Possible solution: remove the react hooks to avoid inconsistencies.

## Tests / Screenshots

-   Help Chat: 
image

-  React in the gatsby-browser.tsx
image


Brenosalv:
Desktop:
image

Mobile:
image

Brenosalv:
Desktop report with the help chat placeholder:
image

gatsby-browser.tsx
```diff
@@ -1,50 +1,65 @@
...
+      {shouldLoadHelpChat && (
+        <Script
+          id="ze-snippet"
+          key="gatsby-plugin-zendesk-chat"
+          strategy={'idle'}
+          async
+          defer
+          src={`https://static.zdassets.com/ekr/snippet.js?key=${ZD_KEY}`}
+        />
+      )}
```
travjenkins, line 46: This is not the solution I'd recommend. I am thinking this should be handled more like how this plugin can load in HubSpot. It does not need to be done exactly like this - but should be loaded from within the component itself.

https://github.com/escaladesports/react-hubspot-form/blob/d439bb88631eb8864e96397a0d4c23e81cf1929f/src/index.js#L44

travjenkins (changes requested):
Do not need to load dynamically

Brenosalv:
@travjenkins I wouldn't consider adding states here since it could cause unnecessary rerenders. Intead we just need to check the screen width and height on the first render and load it or not. Don't you aggree?

Brenosalv:
@travjenkins I think we don't need the component because we don't need the state. We just need to check if the screen width or height is less than 768px on load/first render and we can do it without states. Please let me know if you aggree.

Issue #231, opened 2024-03-19
Add eslint to the build process


travjenkins:
## Description
Need to add eslint to the build process so that it will fail if there are issues. This means we also need to fix all the current issues with this work.

continues work started with https://github.com/estuary/marketing-site/issues/221

## Work required
- [x] add eslint to the build process
- [x] fail build if eslint has errors

## Figma / Designs
N/A

PR #320, opened 2024-05-23
Brenosalv/feature/add-eslint-to-the-build-process

Brenosalv:
#231 

## Changes

-   Fix all the eslint errors and warnings
-   Integrate Eslint checking to the build pipeline


plugins/estuary-rehype-transformers/src/index.js
```diff
@@ -72,6 +72,7 @@ const transform = async ({ htmlAst, htmlNode, getNode }, opts) => {
                         });
                     }
                 }
+                return undefined;
```
travjenkins, line 75: Why was this added?

plugins/estuary-rehype-transformers/src/index.js
```diff
@@ -7,7 +7,7 @@ import rehypeSlug from 'rehype-slug';
 import rehypeToc from 'rehype-toc';
 import { unified } from 'unified';
 import * as visit from 'unist-util-visit';
-import ImgSharpInline from './ImgSharp';
+import ImgSharpInline from './ImgSharp.jsx';
```
travjenkins, line 10: What drove this? I don't think I've ever seen this needed.

gatsby-node.ts
```diff
@@ -92,9 +92,9 @@ export const createPages: GatsbyNode['createPages'] = async ({
         }
     `);
 
-    const allCaseStudies = caseStudyPages.data.allStrapiCaseStudy.nodes;
+    const allCaseStudies = caseStudyPages.data?.allStrapiCaseStudy.nodes;
 
-    allCaseStudies.forEach((node) => {
+    allCaseStudies?.forEach((node) => {
```
travjenkins, line 97: We should really throw an error if this is empty so the build fails and we can catch issue with data missing that was expect.

travjenkins: This applies to a lot of the data below

src/components/Connectors.tsx
```diff
@@ -279,7 +282,7 @@ export const Connectors = ({
 
                                 <StaticImage
                                     src="../images/connectors-bottom.png"
-                                    alt={bottomTitle}
+                                    alt={bottomTitle ?? ''}
```
travjenkins, line 285: We should give this _something_ for an alt.

.github/workflows/deploy-prod.yml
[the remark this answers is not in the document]
Brenosalv: Okay, but shouldn't it be between the ```yarn install``` and the ```yarn build --log-pages```?

```
// ...
- run: yarn install
 - name: Run ESLint
   run: yarn lint
   env:
       CI: true
- name: Run Prettier
  run: yarn format
  env:
      CI: true
- run: yarn run build --log-pages
 // ...
```

gatsby-node.ts
```diff
@@ -372,17 +373,17 @@ export const sourceNodes: GatsbyNode['sourceNodes'] = async ({
     for (const conn of connectors.rows) {
         const usUrl = conn.logo_url?.['en-US'];
         if (!usUrl) {
-            return null;
+            await pool.end();
```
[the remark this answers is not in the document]
Brenosalv: It's necessary because after the ```await pool.end();``` we exit the function with ```return;```. In this case we don't reach the last line of the function.

But I will add a try...finally so we have only one ```await pool.end();``` in the end of the function.

plugins/estuary-rehype-transformers/src/index.js
Brenosalv: Eslint says the function should return a value at the end of arrow function:

```error  Expected to return a value at the end of arrow function  consistent-return```

File: src/utils.ts [hunk 6]
```diff
@@ -1,6 +1,6 @@
 // NOTE: We're assuming that the image name follows the convention of
 // (source|materialization)-name
-const CONNECTOR_IMAGE_RE = /(source|materialize)-([a-z\-]+)/;
+const CONNECTOR_IMAGE_RE = /(source|materialize)-([a-z-]+)/;
```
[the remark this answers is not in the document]
Brenosalv: I removed it because Eslint says it's not necessary and other sources confirmed this. But okay, I will put it back and disable eslint for this.

plugins/estuary-rehype-transformers/src/index.js
Brenosalv: Eslint: ```error  Unable to resolve path to module './ImgSharp'  import/no-unresolved```

plugins/estuary-rehype-transformers/src/index.js
travjenkins: Okay... I think this is probably harmless then.

gatsby-node.ts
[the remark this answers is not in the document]
travjenkins: Oh - now I see. I was thinking this loop was returning something. Now I see it just exits.

I think we should just wrap the code in an `if` that way we always make it to the last line
```
        const fileNode = await createRemoteFileNode({
            url: usUrl,
            createNode,
            createNodeId,
            getCache,
        });

        await createNode({
            connectorId: conn.id,
            logoUrl: usUrl,
            logo: fileNode,
            id: createNodeId(createLogoNodeId(conn.id)),
            internal: {
                type: 'ConnectorLogo',
                contentDigest: createContentDigest(fileNode),
            },
        });
```

gatsby-node.ts
Brenosalv: Makes sense

src/components/CascadingMenu.tsx
```diff
@@ -191,12 +193,12 @@ export const NavMenuList = ({ item }: { item: NavItem }) => {
         >
             <ListItemText
                 primary={
-                    <Link to={item.path} className="global-header-link">
+                    <Link to={item.path ?? ''} className="global-header-link">
```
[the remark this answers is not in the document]
Brenosalv: Yeah, I will make it required.

gatsby-node.ts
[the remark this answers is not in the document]
Brenosalv: Yeah, even better.

plugins/estuary-rehype-transformers/src/index.js
Brenosalv: Yes, harmless.

.github/workflows/deploy-prod.yml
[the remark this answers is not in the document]
travjenkins: Yup - that's where it should go

travjenkins (approved):
lgtm - last round of changes are alright and new build steps ran fine

Issue #313, opened 2024-05-23
Menu and Banner Overlapping


altitude-lucas:
![Image](https://github.com/estuary/marketing-site/assets/167227013/267ba175-6d28-41a9-b88e-072f1eb14855)

PR #326, opened 2024-05-28
Brenosalv/bug/menu-and-banner-overlapping

Brenosalv:
#313 

## Changes

-   Remove overlapping between header and Take a Tour banner

## Tests / Screenshots

image

### DIFF
--- a/src/components/TakeATour/index.tsx
+++ b/src/components/TakeATour/index.tsx
@@ -1,47 +1,48 @@
-import { Button, Container, Link } from './styles';
-import React, { useCallback, useEffect, useState } from 'react';
 import CloseIcon from '@mui/icons-material/Close';
+import IconButton from '@mui/material/IconButton';
+import React, { useCallback, useEffect, useState } from 'react';
+import { Container, Link } from './styles';
 
 const DEFAULTS = {
-  href: '/why',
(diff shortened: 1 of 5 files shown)


Brenosalv:
With banner:
image

Without banner:
image

travjenkins (changes requested):
described in another comment with screenshot

Brenosalv:
> This looks broken on load while waiting for the banner
> 
> image
> 
> 
> 
> This will cause a shift overall too. I think we should just redesign the header to be a bit taller all the time and that allows rooms for the banner. Having anything move/shift ends up hurting the Google performance metrics.
> 
> 

@travjenkins The additional margin top would prevent the header items to be centralized. Another possible solution would be moving the banner to the bottom of the header, out of it. What do you think?

Brenosalv:
> This looks broken on load while waiting for the banner image
> 
> This will cause a shift overall too. I think we should just redesign the header to be a bit taller all the time and that allows rooms for the banner. Having anything move/shift ends up hurting the Google performance metrics.

@travjenkins Do you think it's enough taller?

image

travjenkins (approved):
LGTM

We can skip moving it down as we're redesigning the menu anyway.

PR #328, opened 2024-05-29
Privacy Policy Update : May 28, 2024

travjenkins:
## Changes

- Update privacy policy with latest changes
- moving styling into a shared location

### Privacy
image

### Terms
image

### DIFF
--- a/src/components/LegalWrapper.tsx
+++ b/src/components/LegalWrapper.tsx
@@ -1,0 +1,39 @@
+import styled from 'styled-components';
+
+const LegalWrapper = styled.div`
+  padding-left: calc(min(9%, 146px));
+  padding-right: calc(min(9%, 146px));
+
+  .privacy-underline {
+    text-decoration: underline;
+  }
+
+  .privacy-bold {
+    font-weight: bold;
+  }
+
+  &.privacy-root {
+    .last-updated {
+      font-style: italic;
+    }
+  }
+
+  &.terms-root {
+    display: flex;
+    flex-direction: column;
+
+    h4,
+    h6 {
+      text-align: center;
+    }
+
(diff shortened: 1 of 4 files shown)


src/style.less
```diff
@@ -2553,42 +2553,6 @@ table th {
...
-  }
-}
-
-.terms-last-updated {
-  font-weight: bold;
-  color: #6377ee;
-  margin-top: 0;
-  text-align: center;
-  width: 100%;
-}
```
travjenkins, line 2590: This is the real change

Brenosalv (commented):
I'm currently naming the folders in the components directory according to the corresponding page name as ```Homepage``` folder. This way, I would consider creating a new folder called PrivacyPolicy and including a styles.ts file within it, instead of just having a file named LegalWrapper.tsx directly inside the components folder. What do you think?

src/components/LegalWrapper.tsx
```diff
@@ -0,0 +1,39 @@
+import styled from 'styled-components';
+
+const LegalWrapper = styled.div`
+  padding-left: calc(min(9%, 146px));
```
Brenosalv, line 4: The new horizontal padding we adopted since redesigned homepage is:
```
padding-left: calc(min(10vw, 160px));
padding-right: calc(min(10vw, 160px));
```

Maybe we should adopt the same padding here.

Brenosalv (approved):
Lgtm - Just left two comments.

src/components/LegalWrapper.tsx
travjenkins: I left this as a stand alone due to it being legal docs. I don't _think_ it matters but I didn't want to mess with it.

Issue #314, opened 2024-05-23
Customer Testimonials - Spacing issues


altitude-lucas:
![Image](https://github.com/estuary/marketing-site/assets/167227013/f50fe42f-7d8e-43af-96a4-ce88a35660d2)

![Image](https://github.com/estuary/marketing-site/assets/167227013/03b689dd-8ac7-4884-8ea5-3e226f08e106)

Issue #303, opened 2024-05-20
Replace the carousel with a HTML/CSS one


Brenosalv:
## Description

We should look into getting the carousel switched over to HTML/CSS one: https://css-tricks.com/css-only-carousel/

Here is another article on this: https://dev.to/ngnijland/make-a-responsive-carousel-with-just-css-36ip

There are carousels in the homepage and pricing page.

## Work required

-   [x] Create a component for all the carousels.
-   [x] Replace the homepage carousel with the new carousel component.
-   [x] Replace in the pricing page as well.

PR #335, opened 2024-05-31
Brenosalv/feature/updated-menu-design

Brenosalv:
#303 
#314 
#324 

## Changes

- [x] Add redesigned product menu;
- [x] Add redesigned resources menu

## Tests / Screenshots

-   image
-   image


package.json
```diff
@@ -106,6 +106,7 @@
     "lint:fix": "eslint . --ext .js,.jsx,.ts,.tsx --fix",
     "format": "prettier --check \"**/*.{js,jsx,ts,tsx,json,md}\"",
     "format:fix": "prettier --write \"**/*.{js,jsx,ts,tsx,json,md}\"",
+    "fix": "yarn lint:fix && yarn format:fix",
```
Brenosalv, line 109: @travjenkins I added this script to make fixing things easier before committing, since we don't have the fix in husky.

travjenkins: Makes sense

src/components/HeaderNavbar/CardItem/styles.ts
```diff
@@ -0,0 +1,82 @@
...
+    font-size: 16px;
+    font-weight: 700;
+    line-height: 28px;
+`;
+
+export const Description = styled.p`
+    font-size: 0.875rem;
+    font-weight: 400;
+    line-height: 30px;
+    color: #ffffff;
```
travjenkins, line 48: Could we move the color into a parent to reduce duplication. I know this is a tiny thing but trying to keep CSS as small as possible.

src/components/HeaderNavbar/Resources/index.tsx
```diff
@@ -65,35 +66,34 @@ const HeaderNavbarResources = ({ active, setActive }) => {
...
+                            <StaticImage
+                                src="../../../images/product-tour-2min.png"
+                                alt="Product tour - 2 minutes"
+                            />
+                        </ImageWrapper>
+                        <OutboundLinkOutlined
                             target="_blank"
-                            href={webinarsUrl}
-                            className="cta-button"
+                            href={webinarsUrl} // TODO: replace this URL with the tour video URL
```
travjenkins, line 92: FYI - a `todo` is still in here.

src/globalStyles.ts
```diff
@@ -70,13 +74,30 @@ export const BaseButtonFilledStyling = `
...
+            return lightStyles;
+        case 'dark':
+            return css`
+                ${BaseButtonStyling};
+                color: #ffffff;
+                background-color: transparent;
+                border: 2px solid #5072eb;
+            `;
+        default:
+            return lightStyles;
```
travjenkins, line 98: I think we're fine just having `light` be the default and not checked up above. That way we don't need to store off the lightStyles. 

Also, can we make the border styling shared?

src/images/product-tour-2min.png
travjenkins, line 1: What changed with this?

Brenosalv: I replaced the image with a bigger one.

travjenkins: Cool - asking cause I was thinking "I have seen this exact image before" lol

travjenkins (changes requested):
a few small changes

src/components/HeaderNavbar/Card/styles.ts
```diff
@@ -5,17 +5,27 @@ interface ContainerProps {
...
     position: fixed;
+    left: 50%;
+    transform: translateX(-50%);
+`;
+
+export const MenuBackground = styled.div<ContainerProps>`
+    ${BaseMenuStyling};
+    width: 100vw;
+    height: 100vh;
+    background-color: #00000099;
```
travjenkins, line 18: This looks wrong... there are more than 6 digits. Does hex support opacity with the last two digits?

src/components/header.tsx
```diff
@@ -147,71 +147,73 @@ const Header = (props: { fixedHeader?: boolean }) => {
...
+                            >
+                                Try it free
+                            </OutboundLinkFilled>
+                        </div>
+                    </div>
+                    <div className="global-header-mobile-menu-wrapper">
+                        <button
+                            onClick={() => setMobileMenuOpen((open) => !open)}
+                            className="global-header-mobile-menu-button"
+                            title="Menu"
```
travjenkins, line 209: Can we update this to `Navigation Menu` to be more specific.

Brenosalv: Sure, makes sense.

src/globalStyles.ts
Brenosalv: Makes sense. Will do this.

src/components/HeaderNavbar/Resources/index.tsx
Brenosalv: Yeah, I was keeping it to remove when working on the carousel.

src/components/HeaderNavbar/CardItem/styles.ts
Brenosalv: I can do this but it's a problem that is present in all the css codes. We could create an issue to store all the colors of the entire application in a file as ```globalStylies.ts``` and import from there.

travjenkins: I would like to reduce how much we're importing and reapplying styles if we can. I would like to style a lot of stuff at the top level and then only use styling to override when needed.

travjenkins: And yeah - this is a big problem in the app for sure... this is not coming from the code of just this PR :D

src/components/HeaderNavbar/Resources/items.tsx
```diff
@@ -1,36 +1,63 @@
...
     {
         name: 'Podcasts',
         to: '/podcasts',
-    },
-    {
-        name: 'YouTube',
-        to: 'https://www.youtube.com/watch?v=Ys5BoNqKljc',
+        hasChevronIcon: true,
+        Image: () => (
+            <StaticImage src="../../../svgs/microphone.svg" alt="About us" />
```
travjenkins, line 39: Can we try to get some better `alt` text in here?

travjenkins: Same applies to the two images down below.

Brenosalv: Okay, I will follow the same idea of the About us from the image up above. This way the alt text would be "Podcasts", "Webinars" and "Youtube".

src/components/Carousel/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+                    }
+                }
+            });
+        };
+
+        return new IntersectionObserver(observerCallback, { threshold: 0.5 });
+    }, []);
+
+    React.useEffect(() => {
+        if (hasMounted.current && slideRefs.current[currentSlide]) {
```
Brenosalv, line 67: @travjenkins The scrolling problem here was caused by the ```scrollIntoView()``` being called on first render. Now we are preventing the component to call it initially using a ref.

Brenosalv:
@travjenkins Also, I removed the ```keen-slider``` package and replaced the connectors page carousel with the new one allowing the same colors to the dots.

src/components/Carousel/styles.ts
```diff
@@ -0,0 +1,84 @@
...
+    li {
+        list-style: none;
+        margin: 0;
+        padding: 0;
+    }
+`;
+
+export const Viewport = styled.ul`
+    display: flex;
+    align-items: center;
```
Brenosalv, line 15: https://github.com/estuary/marketing-site/issues/314https://github.com/estuary/marketing-site/issues/314

I resolved the issue by centering the slide content. I believe this is the maximum reduction we can achieve for the spacing between the testimonials and dots from the carousel.

travjenkins: I think this is okay for now. It is better than before in my opinion. Probably see what Lucas thinks.

Eventually we should put some effort into adjusting the height dynamically.

src/components/HeaderNavbar/Card/index.tsx
```diff
@@ -1,14 +1,28 @@
...
+    const isMobile = isRealMobile || isSmallScreen;
+
+    if (isMobile) {
+        return (
+            <Wrapper ref={customRef} {...props}>
+                <Divider />
+                <Container>{children}</Container>
+            </Wrapper>
+        );
+    }
```
travjenkins, line 17: What is driving the need of this? Are we not able to style `MenuBackground` for small screens with just css?

Brenosalv:
@travjenkins As we don't have a mobile version of the redesigned menus, I felt free to develop the responsiveness including an accordion when it's in mobile devices. Also, the mobile version in prod is bugged when clicking on the header links "Product" and "Resources" requiring us sometimes double clicking to open them. Please let me know your thoughts.

src/components/HeaderNavbar/Card/index.tsx
Brenosalv: You're totally right, there is no need of this. Changed to css only.

src/components/Carousel/styles.ts
Brenosalv: I thought about adjusting the height dinamically but I personally didn't like the UX like that.

I talked to Lucas and had a positive feedback about the spacing.

Also, I adjusted the homepage testimonials' logo and font sizes according to figma design. They were too big for mobile.

travjenkins (dismissed):
lgtm

travjenkins (dismissed):
lgtm

travjenkins (dismissed):
lgtm - this **really** makes the nav look better.

travjenkins (approved):
lgtm - hit the preview app and looked around and looked at the changes in the styling (post merge)

PR #338, opened 2024-05-31
Use background image component 

travjenkins:
## Changes

- Switched some items at the bottom of the blog to use the background component to lazy load

## Tests / Screenshots

- No major changes. There is a slight change in where the lines show as the background position is static but seems okay for now.
image

image


Brenosalv (approved):
Lgtm

Issue #273, opened 2024-05-01
Add social media links to Blog posts


altitude-lucas:
Add options to share on various social platforms to the blog template

PR #348, opened 2024-06-10
Brenosalv/feature/add-social-media-links-to-blog-posts

Brenosalv:
#273 

## Changes

- [x] Add "Share this article" section in blog post sidebar
- [x] Replace Slack with Linkedin

## Tests / Screenshots

-   Desktop:
image

-   Mobile:
image


src/templates/blog-post/ShareArticle/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+    const shareMessage = `Check out the article "${articleTitle}"`;
+
+    const articleUrl = window.location.href;
+
+    const articleUrlToClipboard = async () => {
+        try {
+            await navigator.clipboard.writeText(articleUrl);
+        } catch (error) {
+            // Todo: treat this possible error
+        }
```
Brenosalv, line 28: @travjenkins What do you think about using a toast/snackbar/tooltip to show the URL was copied or not, or show an error?

Brenosalv: Displaying an alert does not provide the best UX but maybe is enough.

src/templates/blog-post/ShareArticle/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+                <CopyButton>
+                    <CopyIcon onClick={articleUrlToClipboard} />
+                </CopyButton>
+                <SocialLink
+                    target="_blank"
+                    aria-label="Share article in Linkedin"
+                    href={`https://www.linkedin.com/sharing/share-offsite/?url=${encodeURI(articleUrl)}`}
+                >
+                    <SlackOutlinedIcon />{' '}
+                    {/* TODO: replace this icon with Linkedin one */}
```
Brenosalv, line 44: Waiting Polina to provide me a Linkedin svg icon in Figma.

src/templates/blog-post/ShareArticle/index.tsx
travjenkins: On the dashboard we have the button turn green when it is copied with a tooltip - I think that'll be god (if the buttons are easy to make green)

image
image

src/templates/blog-post/ShareArticle/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+const ShareArticle = ({ articleTitle }: ShareArticleProps) => {
+    const shareMessage = `Check out the article "${articleTitle}"`;
+
+    const articleUrl = window.location.href;
+
+    const articleUrlToClipboard = async () => {
+        try {
+            await navigator.clipboard.writeText(articleUrl);
+        } catch (error) {
+            // Todo: treat this possible error
```
travjenkins, line 27: To handle this error we might want to auto show an input under the icons with the URL.... that way the user could manually copy the path.

src/templates/blog-post/ShareArticle/index.tsx
travjenkins, line 1: The content says "Share article in {provider}". I think it should say "Click to share article on {provider}". Want to make it clear they need to click/interact... also replacing `in` with `on` seems appropriate.

src/templates/blog-post/ShareArticle/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+        }
+    };
+
+    return (
+        <Container>
+            <span>Share this article</span>
```
travjenkins, line 33: Since this is a title of a section I _think_ this should probably be some kind of header. However, I think we'd want to check with Sourabh to be sure.

src/templates/blog-post/ShareArticle/index.tsx
Brenosalv: I like this solution. Thank you!

src/templates/blog-post/ShareArticle/index.tsx
Brenosalv: Yeah, I was not sure about this. I will ask him about this.

gatsby-node.ts
```diff
@@ -260,20 +260,27 @@ export const createPages: GatsbyNode['createPages'] = async ({
...
-                        id: post.id,
-                        previousPostId,
-                        nextPostId,
-                        lastMod: post.updatedAt,
-                    },
-                });
+                if (!post.Slug) {
+                    throw new Error(
+                        `Unable to figure out a slug for the post with id: ${post.id}`
+                    );
```
travjenkins, line 266: I assume we're throwing an errror on this now because we really need the slug line as a description for sharing?

src/templates/blog-post/ShareArticle/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+            setIsCopyFailed(true);
+        }
+    };
+
+    const handleTooltipClose = () => {
+        setTimeout(() => {
+            setIsCopied(false);
+        }, 800);
+    };
+
```
travjenkins, line 50: Can we update this a bit to make the "auto closing" code live next to it

```js
    const copyToClipboard = () => {
        navigator.clipboard.writeText(value).then(
            () => {
                setIsCopied(true);
                setIsCopyFailed(false);
                setTimeout(() => setIsCopied(false), 2000);
            },
            () => {
                setIsCopied(false);
                setIsCopyFailed(false);
            }
        );
    };
```

gatsby-node.ts
Brenosalv: I added this because the build was failing when a slug was undefined. It happened after I change the ```templates``` folder structure to include a ```styles.ts``` file:

```
- const blogPost = path.resolve('./src/templates/blog-post.tsx');
+ const blogPost = path.resolve('./src/templates/blog-post/index.tsx');
```

src/templates/blog-post/ShareArticle/index.tsx
Brenosalv: Done

PR #354, opened 2024-06-14
Integration page quick updates

travjenkins:
## Changes
 - Adding contact us to integrations page
 - Updating the register cta to be styled properly
 - Ran an install and lock updated

### desktop
image

### mobile
image

### DIFF
--- a/src/style.less
+++ b/src/style.less
@@ -1,69 +1,69 @@
 /* CSS Custom Properties Definitions */
 
 :root {
-  --maxWidth-none: 'none';
-  --maxWidth-xs: 20rem;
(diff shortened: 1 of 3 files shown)


src/style.less
```diff
@@ -117,6724 +117,6735 @@ h3,
...
+        margin-bottom: 0px;
 
-  .connector-section-background {
-    background-color: #f9fafc;
-  }
+        button,
+        a {
+            gap: 5px;
+            width: 50%;
+        }
```
travjenkins, line 3440: This is the real change in this file

Brenosalv (approved):
Lgtm

Issue #352, opened 2024-06-14
Menu Bug - Width Error


altitude-lucas:
Bug reported, when menu width is adjusted, many elements in the dropdown disappear: https://www.awesomescreenshot.com/video/28642343?key=e3a5df5df5900fa3beb6abb9c5550837

PR #361, opened 2024-06-19
Brenosalv/bug/menu-bug-width-error

Brenosalv:
#352

## Changes

-    Add missing content to menu on mobile;
-    Add scroll when screen is too small.

## Tests / Screenshots

image

### DIFF
--- a/src/components/HeaderNavbar/Card/styles.ts
+++ b/src/components/HeaderNavbar/Card/styles.ts
@@ -18,7 +18,6 @@ export const Container = styled.div`
     ${globalMaxWidth};
 
     font-family: 'Inter', sans-serif;
-
     display: grid;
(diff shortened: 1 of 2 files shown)


travjenkins (approved):
lgtm

Brenosalv:
> What new content is being shown on mobile? image

@travjenkins Sorry, it was not clear. The missing content was the columns that disappeared when Sourabh zoomed in: https://www.awesomescreenshot.com/video/28642343?key=e3a5df5df5900fa3beb6abb9c5550837

Brenosalv:
> When the window is 1025px wide the hover state of Documents link ends up wrapping. I think we should prevent this from wrapping but have a max width on the item so the text is partially hidden. image

@travjenkins I'll create an issue for this.

Issue #296, opened 2024-05-15
Redesign : Product Page


travjenkins:
## Description

New Product Page is ready to build: https://www.figma.com/design/qhXFvOJPqLgGsr9lsAPYbn/Estuary%3A-Product-Page?node-id=4205-6178&t=NrNvst0KAuxhobVf-1

## Work required

-   [x] Implement all redesigned product page sections

## Figma / Designs

include any screenshots/links of designs

travjenkins:
Getting close. "How It Works" content is in progress. Most other ones are finished

PR #370, opened 2024-06-25
Brenosalv/feature/redesign-product-page

Brenosalv:
#296 

## Changes

-   Add all new section of the redesigned product page.


src/layouts/Product/sections/Hero/index.tsx
```diff
@@ -1,58 +1,97 @@
...
-        <DefaultWrapper>
-            <ContainerHeader>
-                <ContainerImage>
-                    <StaticImage
-                        placeholder="none"
-                        src="../../../../images/lp-product/hero.png"
-                        alt="Screenshot of dashboard"
-                        layout="constrained"
-                        loading="eager"
+        <DefaultWrapperDarkBlue style={{ padding: '50px 0 60px 0' }}>
```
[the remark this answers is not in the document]
travjenkins: That should probably be fine.

src/components/Carousel/index.tsx
```diff
@@ -129,23 +131,17 @@ const Carousel = ({
                         disabled={isLeftArrowDisabled}
                         aria-label="previous testimonial"
```
travjenkins, line 132: I don't think these be hard coded to say `testimonial` anymore.

src/components/Carousel/index.tsx
```diff
@@ -156,7 +152,7 @@ const Carousel = ({
                         disabled={isRightArrowDisabled}
                         aria-label="next testimonial"
                     >
-                        <ChevronRightIcon />
+                        <ChevronRightIcon color={arrowColor ?? '#FFFFFF'} />
```
travjenkins, line 155: Can we make this color and the one up above shared?

src/components/Homepage/SectionOne/index.tsx
```diff
@@ -96,7 +96,7 @@ const SectionOne = () => {
             <MetricCardsList>
                 <MetricCard>
                     <MetricIconWrapper>
-                        <ActiveUsersIcon />
+                        <ActiveUsersIcon color="#FFFFFF" />
```
travjenkins, line 99: Since the white is repeated 3 times let's move it into a `const` up above `SectionOne`

src/components/Product/SectionEight/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+                                    approachable pricing and quick setup to
+                                    deliver change data capture solutions.
+                                </p>
+                            </BottomCaseStudy>
+                        </BottomCaseStudies>
+                    </CaseStudies>
+                }
+                advantages={[
+                    {
+                        id: 7,
```
travjenkins, line 90: It looks like all the advantage should have a unique `id` to keep things safe and not have any overlap. I think we should probably stores these in a single file and then import them as needed. We can still keep 'em grouped together in individual arrays but keeping in a single file will keep the ids from clashing

Something like 

```
const advantagesProduct = [
    {
        id: 7,
        title: '...'
    },
    {
        id: 8,
        title: '...'
    },
    {
        id: 9,
        title: '...'
    },
    {
        id: 10,
        title: '...'
    },
]

const advantagesFoo = [
    {
        id: 11,
        title: '...'
    },
    ...
    {
        id: 14,
        title, '...'
    }
]

const advantagesBar = [
    {
        id: 15,
        title: '...'
    },
    ...
]
```

src/components/Product/SectionEleven/styles.ts
```diff
@@ -0,0 +1,220 @@
...
+
+export const ElasticStreamingEtlLogoWrapper = styled(LogoWrapper)`
+    top: 8.4%;
+    ${commonLeftPositionStyles};
+`;
+
+export const MySQLStreamingEtlLogoWrapper = styled(LogoWrapper)`
+    top: 50%;
+    transform: translateY(-50%);
+    ${commonRightPositionStyles};
```
travjenkins, line 203: This is a spot where we should not use repeated styles within `styled` as it ends up getting repeated a lot. These should be handled with adding classes and keeping the common styling on the parent.

src/components/Product/SectionEleven/styles.ts
```diff
@@ -0,0 +1,220 @@
...
+            background: linear-gradient(to bottom, #6ed5d699, #7284ff99)
+                border-box;
+        }
+    }
+`;
+
+export const SnowflakeStreamingEtlLogoWrapper = styled(LogoWrapper)`
+    top: 50%;
+    transform: translateY(-50%);
+    ${commonLeftPositionStyles};
```
travjenkins, line 181: Same comment as the `commonRight...` stuff

src/components/Product/SectionEleven/styles.ts
```diff
@@ -0,0 +1,220 @@
...
+    gap: 16px;
+    margin: auto;
+
+    @media (max-width: 1280px) {
+        text-align: center;
+    }
+`;
+
+export const StreamingEtlTitle = styled.h3`
+    ${commonH3Styles};
```
travjenkins, line 125: Same as below. These common styles should be handled in the parent.

src/components/Product/SectionFive/index.tsx
```diff
@@ -0,0 +1,198 @@
...
+                                    </TabContent>
+                                ),
+                            },
+                            {
+                                label: 'AI',
+                                content: (
+                                    <TabContent>
+                                        <TabImageWrapper>
+                                            <LogoAndPath tab="ops" />
+                                            <Connectors $tab="ops">
```
travjenkins, line 159: I think these tabs should be `ai` and not `ops`

src/components/Product/SectionFour/styles.ts
```diff
@@ -0,0 +1,68 @@
...
+        color: #5072eb;
+    }
+`;
+
+export const Subtitle = styled.p`
+    font-size: 1.25rem;
+    font-weight: 400;
+    line-height: 30px;
+    text-align: center;
+    color: #47506d;
```
travjenkins, line 43: If the color is gonna be the same between Title and Subtitle I think it could go into the parent?

src/components/Product/SectionTen/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+                        layout="constrained"
+                        loading="lazy"
+                    />
+                }
+                title={
+                    <>
+                        <span>SECURE</span> YOUR DATA
+                    </>
+                }
+                subtitle="Estuary Flow is designed and tested to make sure your data and your systems stay secure. "
```
travjenkins, line 24: Space at the end needs removed
image

src/components/Product/SectionThirteen/styles.ts
```diff
@@ -0,0 +1,138 @@
...
+        font-weight: 700;
+        line-height: 38.4px;
+        margin: 0;
+    }
+
+    @media (max-width: 768px) {
+        font-size: 1.75rem;
+        font-weight: 700;
+        line-height: 33.6px;
+    }
```
travjenkins, line 22: These two media queries have the same max-width... not sure why?

src/components/Product/SectionTwo/Hover/index.tsx
```diff
@@ -0,0 +1,676 @@
...
+                <ConnectorsGroupsLeftWrapper>
+                    <ConnectorsGroup>
+                        <Connector to="/source/oracle-flashback">
+                            <LogoWrapper>
+                                <StaticImage
+                                    placeholder="none"
+                                    alt="Oracle logo"
+                                    src="../../../../images/logos/oracle.png"
+                                    layout="constrained"
+                                />
```
travjenkins, line 113: This is okay for now and needs no changes... we should think of a way to store data for things like this so that all the images, alt, and titles are always in sync. As this is a lot of code that could easily get messed up with a copy/paste.

src/components/Product/SectionTwo/Hover/index.tsx
```diff
@@ -0,0 +1,676 @@
...
+                            </LogoWrapper>
+                            <HiddenLogoTitle>Amazon Kinesis</HiddenLogoTitle>
+                        </Connector>
+                        <Connector to="/source/kafka">
+                            <LogoWrapper>
+                                <StaticImage
+                                    placeholder="none"
+                                    alt="AWS Kinesis logo"
```
travjenkins, line 277: This `alt` text is wrong and should be for Kafka

src/components/Product/SectionTwo/Hover/index.tsx
```diff
@@ -0,0 +1,676 @@
...
+                    </ConnectorsGroup>
+                </ConnectorsGroupsRightWrapper>
+            ) : null}
+            <Fade in={hoveredFlowStepBlock === 'transform'}>
+                {slide === 2 ? (
+                    <div>
+                        <FlowStepImageWrapper $slide={slide}>
+                            <StaticImage
+                                placeholder="none"
+                                alt="Graphic - Flow Step Transform"
```
travjenkins, line 461: I think `Graphic` can be removed from these `alt` texts

src/components/Product/SectionTwo/index.tsx
```diff
@@ -0,0 +1,203 @@
...
+                    hasArrow
+                    aria-label="How Estuary Flow works carousel"
+                    arrowColor="#47506D"
+                >
+                    <Slide>
+                        <DesktopImageWrapper>
+                            <Hover slide={1}>
+                                <StaticImage
+                                    placeholder="none"
+                                    alt="Graphic - Capture"
```
travjenkins, line 43: Again should be able to remove `graphic` from the `alt`

src/images/logos/amazon-s3.png
travjenkins, line 1: What is driving saving off these logos? These are already available from the connectors table... however I guess getting those URLs could be tough right now. I think we're okay right now but should look into using the standard ones... then we get the caching benefit.

src/images/product-page/section-two/desktop/ai-connectors-group.png
travjenkins, line 1: I don't think these are needed anymore due to the tooltip approach.

src/components/Product/SectionTwo/Hover/index.tsx
```diff
@@ -0,0 +1,676 @@
...
+                        >
+                            <FeatureWrapper>
+                                <AiIconComponent color="#5072EB" />
+                                <h4>AI</h4>
+                            </FeatureWrapper>
+                        </HtmlTooltip>
+                    </ConnectorsGroup>
+                </ConnectorsGroupsRightWrapper>
+            ) : null}
+            <Fade in={hoveredFlowStepBlock === 'transform'}>
```
travjenkins, line 455: We should not be duplicating Fade here. We just need a single one and then within it it can control what to show or hide. Otherwise there is way too many duplicate elements.

src/components/Carousel/index.tsx
Brenosalv: You're right. Changed to "previous slide" and "next slide".

src/components/Carousel/index.tsx
Brenosalv: Yes, in fact we should inicialize the arrowColor prop with this value.

src/components/Product/SectionEight/index.tsx
Brenosalv: It's a great idea. Thanks, man!

src/components/Product/SectionEleven/styles.ts
Brenosalv: It makes a lot of sense. I will pay more attention to it next time. Thanks!

src/components/Product/SectionTen/index.tsx
Brenosalv: Wow, did you noticed this in the code or used something to scan and find this?

src/components/Product/SectionThirteen/styles.ts
Brenosalv: Sorry, it shouldn't be there. Removing...

src/components/Product/SectionTwo/Hover/index.tsx
Brenosalv: Yes but since the ```StaticImage``` component only accept constant values from the same file for the props like ```src``` and ```alt```, we can't import the data from elsewhere if it's what you mean.

src/components/Product/SectionTwo/Hover/index.tsx
Brenosalv: Thanks man!

Brenosalv: Copy/paste effect. Thanks, man!

src/components/Product/SectionTwo/Hover/index.tsx
Brenosalv: Okay, I think the word "graphic" won't help us in SEO. Removing...

src/images/logos/amazon-s3.png
Brenosalv: Because the logos for the connectors on the integrations page differ from those in the product page section two diagram. Later we could either replace them with the ones from the integrations page or update the logos on the integrations page to ensure consistency.

src/components/Product/SectionTwo/Hover/index.tsx
Brenosalv: Actually we need multiple Fade components because each piece of the circle requires its own fade effect. If we use only one Fade, once the mouse enters that Fade, all the segments inside it will be faded in, and subsequent hover events will not trigger the fade effect for the individual pieces. Please let me know if it's clear what I mean.

src/components/Homepage/SectionOne/index.tsx
```diff
@@ -60,6 +60,8 @@ const SectionOne = () => {
         }
     `);
 
+    const metricIconColor = '#FFFFFF';
```
travjenkins, line 63: Sorry to be a pain... tiny change but can you declare this outside of the function.

src/components/Product/SectionTen/index.tsx
travjenkins: I just happen to notice it.

src/images/logos/amazon-s3.png
travjenkins: That's good to know... yeah those should be the same for sure. I'll open an issue for us to look into it eventually.

src/components/Product/SectionTwo/Hover/index.tsx
travjenkins: Oh... because these fades are not just controlling showing/hiding but also positioned for triggering?

Brenosalv: Exactly

src/components/Homepage/SectionOne/index.tsx
Brenosalv: Sure, makes sense.

travjenkins (approved):
new changes lgtm

PR #371, opened 2024-06-25
Brenosalv/bug/369-address-robs-comments-in-figma-home-page

Brenosalv:
#369

## Changes

-   Change homepage meta description;
-   Change homepage title and first sentence from section one and two.

Obs: See Rob's comments in Figma: https://www.figma.com/design/qhXFvOJPqLgGsr9lsAPYbn/Estuary%3A-Product-Page?node-id=3627-6071&t=y72P8PH2a7xpikmd-0

### DIFF
--- a/src/components/Homepage/SectionTwo/index.tsx
+++ b/src/components/Homepage/SectionTwo/index.tsx
@@ -49,7 +49,9 @@ const SectionTwo = () => {
                     <span>Move </span>
                     <span>and </span>
                     <span>transform </span>
-                    <span>data with the only CDC platform built for </span>
+                    <span>
+                        data in real-time with the only platform built for{' '}
+                    </span>
                     <span>ops</span>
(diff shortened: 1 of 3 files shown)


src/components/Homepage/SectionTwo/index.tsx
```diff
@@ -49,17 +49,19 @@ const SectionTwo = () => {
                     <span>Move </span>
                     <span>and </span>
                     <span>transform </span>
-                    <span>data with the only CDC platform built for </span>
+                    <span>
+                        data in real-time with the only platform built for{' '}
```
travjenkins, line 53: I think we can remove this extra space here

Brenosalv: We don't see it before eslint adds it as {' '} haha. Thanks for pointing that.

PR #374, opened 2024-06-27
letting their be numbers in the connector image name

travjenkins:
## Changes

- Update the regex to handle connector images with numbers in them. This is currently only needed for `S3` related connectors

### All the connectors with `S3` in the name
image

image

image

image


Brenosalv (approved):
Lgtm

PR #378, opened 2024-07-03
Updating meta description

travjenkins:
## Changes

- Hard coded the new content into the header for right now. 
- Opened up https://github.com/estuary/strapi-admin/pull/13 so that we can just move this into Strapi

### Fivetran
image

### Confluent
image

### Airbyte
image

### Debezium
image


Brenosalv (approved):
Lgtm

PR #385, opened 2024-07-08
Brenosalv/bug/product-page-section-two-messed-up-on-safari

Brenosalv:
## Changes

-   Align connectors inside the pills

## Tests / Screenshots
image


travjenkins (approved):
lgtm - the issues called out are a lot better than what I was seeing initially.

Issue #334, opened 2024-05-30
Redesign : Integration Page


travjenkins:
## Description

describe what you would like changed

## Work required

-   [ ] engineering will fill this in

## Figma / Designs

include any screenshots/links of designs

PR #389, opened 2024-07-09
Brenosalv/feature/334-redesign--integration-page

Brenosalv:
#334 

## Changes

-   Add new redesigned sections of the integration page.
-   Consider using LESS css instead of styled-components so we can solve the className mismatch issue

## Figma

-   https://www.figma.com/design/qhXFvOJPqLgGsr9lsAPYbn/Estuary%3A-Product-Page?node-id=5690-16077&t=6L2tiDBicq0Kq7XT-0


types.d.ts
```diff
@@ -6,3 +6,5 @@ declare module '*.svg' {
 declare module '*.png' {
     export default string;
 }
+
+declare module '*.module.less';
```
Brenosalv, line 10: @travjenkins When using LESS with CSS modules, the css files are not being found by Typescript when importing the classes. For example:

```
import {
    connectorBackground,
    connectorDescriptionContainer,
    descriptionLight,
    flowImage,
    flowImagesWrapper,
    sectionTitleLight,
    sourceConnectorImage,
    titleAndImages,
} from '../styles.module.less';
```
It returns a Typescript error:
```
Cannot find module '../styles.module.less' or its corresponding type declarations. ts(2307)
```

Adding this declaration ```declare module '*.module.less';``` makes the error go away but still not sure if it's the best practice or if it should be declared differently.

src/components/Integration/SectionOne/styles.module.less
Brenosalv, line 1: @travjenkins The css code with LESS and CSS modules approach is being implemented like that, recommended by the plugin and it's working properly.

src/components/Integration/SectionOne/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+    const sourceConnectorLogo =
+        sourceConnector.logo.childImageSharp?.gatsbyImageData &&
+        getImage(sourceConnector.logo.childImageSharp.gatsbyImageData);
+    const destConnectorLogo =
+        destConnector.logo.childImageSharp?.gatsbyImageData &&
+        getImage(destConnector.logo.childImageSharp.gatsbyImageData);
+
+    return (
+        <section className={defaultWrapperDarkBlue}>
+            <div className={mainContent}>
```
Brenosalv, line 49: @travjenkins And then the classes are imported like this and set to the classNames. Please check lines 14 to 30 and line 48 and 49 for example.

travjenkins: Okay... so with this approach I would probably not have so many class names and use some nested selectors so we don't have to import _so much_.

Brenosalv: @travjenkins Yes, I tried to do this but it requires to import the nested classNames as well.

src/images/integration/section-one/integration-bg.png
Brenosalv, line 1: I replaced the image in section one, originally 198KB in size, with CSS, resulting in an 188KB reduction in file size. The CSS file itself is 10KB.

src/components/Integration/SectionFive/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+    buttonsWrapper,
+    container,
+    description,
+    overlayVectorLeft,
+    title,
+} from './styles.module.less';
+
+const SectionFive = () => {
+    return (
+        // TODO: Replace this overlay with BackgroundImageWrapper - How to add this new background image file to storage?
```
travjenkins, line 18: Please go ahead and remove the background image for now. This makes the pages a bit quicker and we can add it back in later after the image loading performance fixes are implemented.

src/components/Integration/SectionOne/index.tsx
```diff
@@ -63,56 +86,92 @@ const SectionOne = ({ sourceConnector, destConnector }: SectionOneProps) => {
...
+                            <input
+                                placeholder="Email"
+                                type="email"
+                                required
+                                className={emailInput}
+                                value={email}
+                                onChange={handleEmailChange}
+                            />
+                            {/* TODO: Move this button style to LESS */}
+                            <ButtonFilled
```
travjenkins, line 115: This button should have the `pointer` styling.

src/components/PricingCalculator/BrandCard.tsx
travjenkins, line 1: I _think_ this is somewhere in the Pricing component but not sure.

Seeing this locally
image

src/components/Integration/SectionOne/index.tsx
Brenosalv: Yes, it was fixed already. it's in the build that is currently running.

src/components/PricingCalculator/BrandCard.tsx
Brenosalv: Ah it always happens when setting a styled component prop without the "$" as prefix. I will fix this.

types.d.ts
travjenkins: As long as everything builds and works in the IDE I am good. I am not 100% sure on the best practices.

src/images/estuary-icon-big.png
travjenkins, line 1: We already have a large Estuary icon. I think we could use that one.

image

src/components/Integration/advantages.tsx
```diff
@@ -0,0 +1,30 @@
+export const sectionSixAdvantages = [
+    {
+        id: 1,
```
travjenkins, line 3: I think this is fine for now... especially if we think we might end up with more custom `id` values. However, if we do not end up doing that we should just use `index` in the loop to generate a unique key.

src/components/Integration/SectionOne/index.tsx
```diff
@@ -0,0 +1,203 @@
...
+                    icon={<LatencyIcon color={metricIconColor} />}
+                    value="&#60;100MS"
+                    label="End-to-end latency"
+                />
+                <MetricCard
+                    icon={<SingleDataflowIcon color={metricIconColor} />}
+                    value="7+GB/SEC"
+                    label="Single dataflow"
+                />
+            </div>
```
travjenkins, line 197: Idea for the future.

We might want to start always making these kind of "list components" stand alone components and then always reuse them. Since they seem to be repeated design elements it could help make sure everything is kept in sync and we can easily find all the content in a single place.

travjenkins (commented):
A few small callouts.

src/images/estuary-icon-big.png
Brenosalv: Tried to find it and had no success before you show me. Thanks!

src/components/Integration/advantages.tsx
Brenosalv: As the list of advantages are static, it really makes sense to use ```ìndex``` instead of ```id```.

Brenosalv: For the product page the id is necessary but for the integration page, since it has only one section with advantages, the index is enough. So I will make the id optional and use index if the id is not set.

src/components/PricingCalculator/styles.ts
```diff
@@ -10,9 +10,9 @@ import { ButtonFilled } from '../../globalStyles';
 type Theme = { $isDarkTheme: boolean };
 
 export const PricingSlider = muiStyled(Slider)(({
-    isDarkTheme = false,
+    isdarktheme,
 }: {
-    isDarkTheme?: boolean;
+    isdarktheme?: string;
```
travjenkins, line 15: Can you update this type to

```
isdarktheme?: "true" | "false"
```

travjenkins (approved):
lgtm

PR #390, opened 2024-07-10
Update the blog date information

travjenkins:
## Changes

- Added data to schema
- Added data to be visible
- Marked `Published` and `Updated` when both are shown

### Updated with no tags
image

### Updated with tags
image

### No Update no tags
image

Schema Updated
image


Brenosalv (approved):
Lgtm

Issue #278, opened 2024-05-02
H3 - Add Numbering in Blog Posts


altitude-lucas:
After merging the blog post template, it seems that we lost the number formatting in H3's - This is on the right side of the blog page (Expand "In this Article"). E.g [https://estuary.dev/braze-to-bigquery/](https://www.google.com/url?q=https://estuary.dev/braze-to-bigquery/&sa=D&source=docs&ust=1714643916294076&usg=AOvVaw0P8w4VycZw1eD5hYXoY8NB).

![Image](https://github.com/estuary/marketing-site/assets/167227013/605b45fe-eebf-454a-90f9-aad0d546cf6f)

![Image](https://github.com/estuary/marketing-site/assets/167227013/5e29df4e-b31a-45d9-9301-b4a0d81cf9cf)

Brenosalv:
@altitude-lucas Actually it's reflecting the current figma layout. There are no numbers as in the previous version:

In Figma - https://www.figma.com/file/qhXFvOJPqLgGsr9lsAPYbn/Estuary%3A-Product-Page?type=design&node-id=0-1&mode=design&t=H2wUyozeOMoxLmNN-0

![Image](https://github.com/estuary/marketing-site/assets/60396753/fc829c1e-1b24-41f3-a8cd-f8db768eb45c)

In production - [https://estuary.dev/braze-to-bigquery/](https://www.google.com/url?q=https://estuary.dev/braze-to-bigquery/&sa=D&source=docs&ust=1714643916294076&usg=AOvVaw0P8w4VycZw1eD5hYXoY8NB):

![Image](https://github.com/estuary/marketing-site/assets/60396753/9ba3595e-b95d-4d63-99c7-408ba470a616)

Should we really add the numbers back?

altitude-lucas:
@Brenosalv yes, this helps with indexing and SEO - please add the numbers back to specifically H3's

SourabhEstuary:
@Brenosalv @travjenkins Could we make the table of contents expand or highlight only the section we're currently reading? It'd make navigation much smoother for users. Also, I think reducing the font size of the table of contents section would help to showcase the entire TOC at a glance.

![image](https://github.com/user-attachments/assets/75874f9a-66ef-4c96-a8e2-8fa23719c7c6)

travjenkins:
That is how it works. What exactly do you want to have happen?

Brenosalv:
I agree with reducing the font size.

[...]

PR #391, opened 2024-07-10
Get Blog Post TOC rendering with numbers

travjenkins:
## Changes

https://github.com/estuary/marketing-site/issues/278

- Switched to an `ol`
- Updated styling a bit

## Tests / Screenshots

image

### DIFF
--- a/src/components/BlogPostToc.tsx
+++ b/src/components/BlogPostToc.tsx
@@ -14,30 +14,32 @@ type TocItem = {
 
 type RenderTocItemProps = {
     item: TocItem;
+    selectKey: string;
+    selectedItem: string | null;
     depth: number;
     handleItemClick: (id: string) => void;
-    isSelected: boolean;
 };
 
 const RenderTocItem = ({
     item,
     depth,
     handleItemClick,
-    isSelected,
+    selectKey,
+    selectedItem,
 }: RenderTocItemProps) => {
-    if (depth > 1) {
-        return null;
-    }
+    console.log('selectKey', selectKey);
+    console.log('selectedItem', selectedItem);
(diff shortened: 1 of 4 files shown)


src/components/BlogPostToc.tsx
```diff
@@ -138,7 +137,7 @@ export const RenderToc = ({ items }: { items: TocItem[] }) => {
                     </Typography>
                 </AccordionSummary>
                 <AccordionDetails className="accordion-side-padding">
-                    <ul>
+                    <ol role="list">
```
travjenkins, line 140: This will mark with numbers but also I feel like a more appropriate tag to use semantically.

Brenosalv (approved):
Lgtm

Brenosalv:
In my commit, I made the following changes:

- Set the accordion to be expanded by default.
- Set the same line height for both main and nested items.
- Replaced bullet points with numbers.
- Included nested items to update the selected item on scrolling.
- Added an underline with a 4px offset to all link items, not just nested ones.
- Changed the text color of items to black instead of default link blue color.
- Removed the extra margin from the accordion summary.
- Migrated <li> styles to classes .tocItem and .isItemSelected.

Brenosalv:
The sections do not have a wrapper, so it's hard to detect if they are a little bit or half into the section or not. That's why it's changing the selected link from TOC when the heading shows up.

Instead of:

```
<section id="section-1">
<h4>
<p>
<p>
<p>
</section>
<section id="section-2">
<h4>
<p>
<p>
<p>
</section>
...
```

We have:
```
<h4 id="section-1">
<p>
<p>
<p>
<h4 id="section-2">
<p>
<p>
<p>
...
```

PR #394, opened 2024-07-11
Improve performance for blog posts

travjenkins:
Removing linting/prettier from patches and plugins

### Server Client mismatch
- Removing the `SSR` setting as that just hides the problem
- Adding more default options when starting dev server to catch possible issues
- Added some patch to the inline images so that they would set `0px` instead of `0` as that was getting picked up in the mismatch
     - image
 

### Misc
- Remove `plugin` and `patches` from lint and prettier

#### Prod 
image

#### Local - no longer adds extra random values
image

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,5 @@
 node_modules
 dist
 public
+plugins
+patches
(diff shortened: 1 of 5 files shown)


Brenosalv (approved):
Not tested yet but the code looks good to me.

Issue #395, opened 2024-07-12
Change meta og:Image and Twitter:image Given Pages


SourabhEstuary:
**Current Status**
Currently we are using Estuary logo for preview image on all social channels.

**Requirement**
update meta Og:Image and Twitter:image for below-given pages:

### Future PR (`open_graph_image` added via https://github.com/estuary/strapi-admin/pull/14)
- [ ] https://estuary.dev/vs-fivetran/
Add this image: https://estuary.dev/static/c1f82a30997106309b458c1cc8b05fb1/a1729/fivetranvestuary3_1_9eba281eac.webp

- [ ] ~https://estuary.dev/vs-airbyte/~
~Add this image:~
~https://estuary.dev/static/886e3a65dce4bef66f4ecb74d4debde8/719cf/Group_22561_d679a10970.webp~

**Best recommended size for OG:image and Twitter:image :1200x630** 

### Completed https://github.com/estuary/marketing-site/pull/398
- [x] https://estuary.dev/
Add this image: https://estuary.dev/static/078d75f95a6a1f1f2b01bbb7ef6aa6d8/a0375/flow-desktop.webp

- [x] https://estuary.dev/product/
Add this image: https://estuary.dev/static/4b50c1031e04dcfc3d4166f32c787d36/ffa0c/real-time-graphic.webp

- [x] https://estuary.dev/pricing/
Add this image: https://estuary.dev/static/1699becc9a7cfb450f7e2fa4f094b50a/c7a5c/graphic_parent.webp

travjenkins:
For the `vs` pages we'll need to add this to Strapi as those are auto generated.

travjenkins:
@altitude-lucas - I opened this up this morning https://github.com/estuary/strapi-admin/pull/14

After Joseph is back he can review and then deploy the changes to Strapi.

travjenkins:
Closing - we have completed everything that we need/can for now.

The compare pages are changing with https://github.com/estuary/marketing-site/issues/457

PR #398, opened 2024-07-12
Updating SEO images

travjenkins:
## Changes

https://github.com/estuary/marketing-site/issues/395

- Product, Home, and Pricing

### Product
image

### Home
image

### Pricing
image

### DIFF
--- a/src/layouts/Pricing/Head/index.tsx
+++ b/src/layouts/Pricing/Head/index.tsx
@@ -1,12 +1,26 @@
 import React from 'react';
 
+import { graphql, useStaticQuery } from 'gatsby';
 import Seo from '../../../components/seo';
 
 export const Head = () => {
+    const { metaImg } = useStaticQuery(graphql`
+        query {
+            metaImg: file(relativePath: { eq: "pricing/graphic_parent.png" }) {
+                childImageSharp {
+                    fixed(width: 400) {
+                        src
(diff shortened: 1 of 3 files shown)


src/layouts/Pricing/Head/index.tsx
```diff
@@ -1,12 +1,26 @@
...
+        query {
+            metaImg: file(relativePath: { eq: "pricing/graphic_parent.png" }) {
+                childImageSharp {
+                    fixed(width: 400) {
+                        src
+                    }
+                }
+            }
+        }
+    `);
```
Brenosalv, line 17: Since only the ```eq``` changes, can we reuse this code in the other two queries?

travjenkins: There might be but I am not familiar with GraphQL and it complains a lot when I try to share the query. If you know how please update it.

travjenkins: I think we'd have to add this to the page context... might be best to handle that along when we figure out how we'll get the `vs` pages wired up.

Brenosalv: Oh I thought sharing the query would be simple. Let's keep it that way then.

Brenosalv (approved):
Lgtm

PR #403, opened 2024-07-16
Try to get blog post images to not cause server mismatch issues

travjenkins:
## Changes

-   changes...

## Tests / Screenshots


plugins/estuary-rehype-transformers/package.json
travjenkins, line 1: We should try to make sure this version stays in synce with the main branch. This being different caused _some_ (sadly not all) of the mismatches.

PR #406, opened 2024-07-17
Updating image plugin version

travjenkins:
## Changes

- Updating version of `gatsby-plugin-image` in the rehype
- Recreated patch for new version as without it the images won't show up in the blog posts

## Tests / Screenshots

- N/A


Brenosalv (approved):
Lgtm

Issue #203, opened 2024-02-23
Fix blog post image alt text


jshearer:
**Bug**
The images inside of a blog post have the default alt text `Blog Post image`

**Expected**
The alternative text should come from the Strapi response

**Screenshots**
Example:
Renders with the incorrect `alt` tag
![image](https://github.com/estuary/marketing-site/assets/270078/3aeff20f-7b2b-419b-a6b2-70e0fa04260f)

The text is set in Strapi
![image](https://github.com/estuary/marketing-site/assets/270078/cc1e9961-ae72-4bec-a228-9707d471e676)

The proper default is returned in the `Body`
The default is returned in the `childHtmlRehype`
![image](https://github.com/estuary/marketing-site/assets/270078/3fb94946-1b87-4253-b554-65fac68f18a8)

travjenkins:
We are missing the `alt` text somewhere as that is the fallback value.

![image](https://github.com/estuary/marketing-site/assets/270078/ec2cfd8f-b004-461b-b291-ec1af3c08340)

Brenosalv:
I think it's something related to the patch ```@draftbox-co+gatsby-rehype-inline-images+1.0.2.patch```. Someway it's not setting the ```alt``` to the ```imgOptions``` properly.

PR #408, opened 2024-07-17
Brenosalv/bug/203-fix-blog-post-image-alt-text

Brenosalv:
#203 

## Changes

-   Fix image alt texts from blog posts being undefined/null

## Tests / Screenshots
http://localhost:8000/avoid-mds-real-time-tax/:
image

### DIFF
--- a/gatsby-config.ts
+++ b/gatsby-config.ts
@@ -179,7 +179,7 @@ const cfg: GatsbyConfig = {
                 // Plugins configs (optional but most likely you need one)
                 plugins: [
                     {
-                        resolve: `gatsby-rehype-inline-images`,
+                        resolve: '@draftbox-co/gatsby-rehype-inline-images',
                         // all options are optional and can be omitted
                         options: {
                             // all images larger are scaled down to maxWidth (default: maxWidth = imageWidth)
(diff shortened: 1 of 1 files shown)


gatsby-config.ts
```diff
@@ -179,9 +179,9 @@ const cfg: GatsbyConfig = {
                 // Plugins configs (optional but most likely you need one)
                 plugins: [
                     {
-                        resolve: '@draftbox-co/gatsby-rehype-inline-images',
+                        resolve: `gatsby-rehype-inline-images`,
```
travjenkins, line 182: If we changed this should we also have a package.json update?

Brenosalv: Actually it's not the best solution since it's not converting the img to GatsbyImage as before.

The real problem here is that the alt is not present in the img-sharp-inline element. I'm trying to understand how to add it.

travjenkins: Yeah - this one is probably going to be a bit tricky to figure out - but super valuable.

Brenosalv:
I'm thinking it's something related to the patch ```@draftbox-co+gatsby-rehype-inline-images+1.0.2.patch````. Someway it's not setting the ```alt``` to the ```imgOptions``` properly.

PR #421, opened 2024-07-30
Adding SSO to pricing list

travjenkins:
## Changes

- Adding key features to pricing page
- Cleaning up old pricing tab files that seem to no longer be used

## Tests / Screenshots

image


src/components/PlanTabs.tsx
travjenkins, line 1: I cannot see where this is being used so removing. Hoping if it is used it'll break build.

Brenosalv (approved):
Lgtm

PR #428, opened 2024-08-05
Fix blog post image alt text

Brenosalv:
#203 

## Changes

-   Set missing alt text to formatted image element in package patch.

## Tests / Screenshots

image

### DIFF
--- a/patches/@draftbox-co+gatsby-rehype-inline-images+1.0.2.patch
+++ b/patches/@draftbox-co+gatsby-rehype-inline-images+1.0.2.patch
@@ -1,5 +1,5 @@
 diff --git a/node_modules/@draftbox-co/gatsby-rehype-inline-images/gatsby-node.js b/node_modules/@draftbox-co/gatsby-rehype-inline-images/gatsby-node.js
-index 320309d..f5d6444 100644
+index 320309d..2a9fc55 100644
 --- a/node_modules/@draftbox-co/gatsby-rehype-inline-images/gatsby-node.js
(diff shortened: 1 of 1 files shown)


patches/@draftbox-co+gatsby-rehype-inline-images+1.0.2.patch
```diff
@@ -31,16 +31,29 @@ index 320309d..f5d6444 100644
...
+   const url = node.properties.src;
++
+   let imageNode;
+ 
+   try {
+@@ -117,6 +113,7 @@ const replaceNewImage = async (node, cache, store, createNode, createNodeId, rep
+   if (!imageNode) return;
+   let formattedImgTag = {};
+   formattedImgTag.url = url;
++  formattedImgTag.alt = node.properties.alt;
```
Brenosalv, line 52: Real change.

travjenkins: This is such a small change but hopefully has a huge impact for the blog SEO!

Awesome man.

Brenosalv: Finally... just one line! 🙌🏻

travjenkins: Oh man... I am silly... I totally didn't pay attention that only the one line had the `++`.

travjenkins (approved):
LGTM - I tested preview and looks alright.

I guess there is a change that the `node.properties` might be empty sometimes but that should be okay since we are already missing the `alt` already.

Issue #439, opened 2024-08-06
Contact Us Page


altitude-lucas:
Create a contact us page that’s almost identical to [this](https://hevodata.com/contact/) and link to it in our footer (use the form from https://estuary.dev/about/#contact-us)

Address is 244 5th Ave, Suite 1277, New York, NY 10001

Brenosalv:
Sure, I can do this.

PR #443, opened 2024-08-07
Contact us page

Brenosalv:
#439 

## Changes

-   Create contact us page.

## Tests / Screenshots

-   Reference: https://hevodata.com/contact/

### DIFF
--- a/src/components/ContactUs/SectionTwo/index.tsx
+++ b/src/components/ContactUs/SectionTwo/index.tsx
@@ -1,0 +1,32 @@
+import React from 'react';
+import BuildingIcon from '@mui/icons-material/Apartment';
+import {
+    section,
+    container,
+    iconWrapper,
+    title,
+    card,
+    city,
+    address,
+} from './styles.module.less';
+
+const SectionTwo = () => {
+    return (
+        <section className={section}>
+            <div className={container}>
+                <div className={iconWrapper}>
+                    <BuildingIcon htmlColor="#5072EB" fontSize="large" />
+                </div>
+                <h2 className={title}>OUR OFFICE</h2>
+                <div className={card}>
+                    <h3 className={city}>NEW YORK</h3>
+                    <p className={address}>
+                        244 5th Ave, Suite 1277, New York, NY 10001
+                    </p>
+                </div>
+            </div>
+        </section>
+    );
+};
+
+export default SectionTwo;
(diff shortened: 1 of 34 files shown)


Brenosalv:
Requested Sourabh to provide me the title and meta description of the contact us page.

src/pages/about.tsx
travjenkins, line 1: Since so many pages link to here I think either this form needs to stay OR we then need to add a little section that allows the user to click on something to be taken to the new page.

@altitude-lucas ... I think you can make the final call here.

src/components/ContactUs/SectionTwo/index.tsx
travjenkins, line 1: Nit pick - can we share this information between these two places. That way we don't have to worry about them getting out of sync.

image

Brenosalv:
> @Brenosalv / @altitude-lucas - we will also nee to make a call if we keep the HubSpot modal opening buttons or have those link to this page (probably open them in a new page). This could also easily be something we decide later as I think the modal experience is fairly smooth.

Redirecting to the Contact Us page give users the opportunity to view the address and subscribe to the newsletter after submitting the initial form. However, using a modal is also an effective option.

src/components/Homepage/SectionOne/styles.ts
```diff
@@ -70,27 +70,17 @@ export const HomepageHeader = styled.div`
...
-        font-size: 4.5rem;
-
-        @media (max-width: 768px) {
-            line-height: 43.2px;
-            font-size: 2.25rem;
-        }
+    & > .white-text {
+        color: #fff;
     }
 `;
```
Brenosalv, line 85: Leveraging this PR to replace the nth-child selector with a class for a better performance.

Brenosalv: Also targetting this:
image

Brenosalv: I also reduced the font size of title and description of homepage section one.

Brenosalv:
> @Brenosalv / @altitude-lucas - we will also nee to make a call if we keep the HubSpot modal opening buttons or have those link to this page (probably open them in a new page). This could also easily be something we decide later as I think the modal experience is fairly smooth.

@altitude-lucas Thoughts about the modal?

altitude-lucas:
@Brenosalv I'm not sure what you mean here, can you include a screenshot?

altitude-lucas:
@Brenosalv ah, I understand - let's redirect to the page, this seems like a more standard approach

Brenosalv:
@altitude-lucas Should we keep the modal for these other buttons that are not Contact Us?
"Need More?" button:
image

Section ten from integrations page "button Let Us Show You":
image

altitude-lucas:
@Brenosalv yes, I would leave it for the need more button

Brenosalv:
Leveraged this PR to fix this button hiding text:
image

travjenkins (approved):
lgtm

Issue #429, opened 2024-08-05
Split global CSS code into smaller chunks


Brenosalv:
**Contributes to**
https://github.com/estuary/marketing-site/issues/455

**Tech debt**
Currently the global CSS code `style.less` includes a bunch of non-global classes and other styles. Having a single, large CSS file like style.less with around 5800 lines of code can impact both performance and maintainability.

**Expected**
Implement code splitting by moving each chunk of CSS to their specific folders where they are used and keep the `style.less` with only the global CSS code.

**Screenshots**
It has 5886 lines of code:
<img width="915" alt="image" src="https://github.com/user-attachments/assets/4eac21eb-df74-4b7c-8ced-93d7fc12362d">

commons.css:
<img width="997" alt="image" src="https://github.com/user-attachments/assets/efeeefa1-81d6-4112-825c-4bd9fa59c7fe">
<img width="997" alt="image" src="https://github.com/user-attachments/assets/913adae9-02fa-42a7-a3a8-c8d2a2ba9a0c">

Brenosalv:
Having multiple less css modules is returning these warnings:
<img width="795" alt="image" src="https://github.com/user-attachments/assets/4700230b-7f64-4a3f-8814-2c34335e447f">

It's like webpack does not handle conflicting importing order between the css modules with Gatsby.

Possible solutions:
- Quick 1min read article: https://robertmarshall.dev/blog/fix-warn-chunk-commons-mini-css-extract-plugin-error-in-gatsby-js/
- Issue discussion from Webpack Github: https://github.com/webpack-contrib/mini-css-extract-plugin/issues/250

travjenkins:
I think we are fine with this. We do not use a lot of nesting and specificity in our styling so it should not cause any big issues.

PR #448, opened 2024-08-12
Split global css code into smaller chunks - header

Brenosalv:
#429 

## Changes

-   Split global css code into smaller chunks and move non-global classes to their respective folders.

OBS: Since we decided to merge the code splitting piece by piece, this PR is only for the header.


Brenosalv:
Tests for the header:

Desktop:
Screenshot 2024-08-12 at 13 26 10
image
image


Tablet:
image
image
image

Mobile:
image
image
image

src/components/HeaderNavbar/styles.module.less
```diff
@@ -0,0 +1,41 @@
...
+    width: 100%;
+  }
+}
+
+.menuChevron {
+  transition: transform 150ms;
+  transform: rotate(90deg);
+}
+
+.isActive {
```
travjenkins, line 27: I think we should rename this one. It was pretty confusing to see an imported class named just like a normal boolean variable.

Maybe... `activeClass` or even just tack on "style" to the end - `isActiveStyle`

travjenkins (changes requested):
1 small change from static review

src/components/HeaderNavbar/styles.module.less
Brenosalv: Changed to `activeHeaderLink`.

travjenkins (approved):
lgtm

Issue #360, opened 2024-06-19
Base case study links in menu on Strapi


travjenkins:
## Description

Want to make the Case Studies links in the header menu based on Strapi content so as new ones are added the site does not need new coding to show them.

## Work required

- [x] Add new data to strapi (https://github.com/estuary/strapi-admin/pull/13)
- [x] Fill in new data (see content linked below)
- [x] Switch site to use Strapi to make links
    - This will need to have a `max` limit coded into it
- [x] Make links based on Slugs
- [x] Design how we'll handle long lists
- [x] Test all case study links work
- [ ] Might need to clean up the slugs that have `casestudy/` in the value

Existing Strapi
![image](https://github.com/estuary/marketing-site/assets/270078/0db97bcb-9adf-40d1-ab05-740179c5f2ae)

## Figma / Designs

include any screenshots/links of designs

## Notes from Rbo
Format should be as follows:
Customer name
1-line description (with link)
Thumbnail centered to the left

Updated text is here:
https://docs.google.com/document/d/1skS66dG4Tn2GTd9A0KcUEYF6yu6C_dy1qsTtBMHLJ3M/edit?usp=sharing

Brenosalv:
@travjenkins will you be responsible for adding the new data to Strapi? I don't have access to the Strapi admin repo.

travjenkins:
Yeah - that is added already and just needs deployed. I need to get with Joseph to know how the deployment works.

travjenkins:
Yeah - that is added already and just needs deployed. I need to get with Joseph to know how the deployment works.

Brenosalv:
Cool - please let me know once it's ready to me.

travjenkins:
@Brenosalv - this is no longer blocked!

Brenosalv:
@travjenkins @altitude-lucas @EstuaryRob We need to review these logos:
![Image](https://github.com/user-attachments/assets/99849f2b-98ec-44fb-99d2-43fe55db7302)

PR #450, opened 2024-08-13
Use case studies from Strapi on links from resources header menu

Brenosalv:
#360 

## Changes

-   Replace current Resources menu case studies data with data coming from Strapi;
-   Set a limit to the number of case studies listed in the Resources menu.

## Tests / Screenshots

image


Brenosalv:
@travjenkins Since it's not in the scope of the issue description, I was focusing only on the menu. But yes, we can do this.

travjenkins (approved):
lgtm

PR #452, opened 2024-08-14
Update pricing to match new approach to charging per connector

travjenkins:
## Changes

High level - we will now charge...
$100 per connectors for the first 6
$50 per connector for 6-12
"talk to us" for over 12 connectors

## Tests / Screenshots

-   Pricing page:
image
image
image

-   Homepage:
image
image
image
image

### DIFF
--- a/src/components/PricingCalculator/shared.ts
+++ b/src/components/PricingCalculator/shared.ts
@@ -1,5 +1,5 @@
 export const selfServiceConnectorLimit = 12;
-export const halfSelfServiceConnectorLimit = 6;
+export const halfSelfServiceConnectorLimit = selfServiceConnectorLimit / 2;
 export const maxConnectors = selfServiceConnectorLimit + 1;
 export const fullPricingPerConnector = 100;
 export const halfPricingPerConnector = fullPricingPerConnector / 2;
--- a/src/components/PricingCalculator/shared.ts
+++ b/src/components/PricingCalculator/shared.ts
@@ -1,0 +1,2 @@
+export const selfServiceConnectorLimit = 12;
+export const maxConnectors = selfServiceConnectorLimit + 1;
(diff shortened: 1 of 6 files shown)


src/utils.ts
```diff
@@ -79,7 +79,7 @@ export const currencyFormatter = Intl.NumberFormat('en-US', {
 });
 
 export const calculatePrice = (gb: number, connectors: number) => ({
-    estuary: costPerGB_calc * gb + 100 * connectors,
+    estuary: costPerGB_calc * gb + (connectors <= 6 ? 100 : 50) * connectors,
```
travjenkins, line 82: I know this is a bit over engineering... but I think we should probably look at storing the `6` and the `100` and `50` off into a shared location so we can update in one spot and have the content update as well.

src/components/PricingCalculator/shared.ts
```diff
@@ -0,0 +1,5 @@
+export const selfServiceConnectorLimit = 12;
+export const halfSelfServiceConnectorLimit = selfServiceConnectorLimit / 2;
```
Brenosalv, line 2: Is it okay to have the `halfSelfServiceConnectorLimit` directly depending on the `selfServiceConnectorLimit`?

Now we are doing this:
```
estuary:
        costPerGB_calc * gb +
        (connectors <= halfSelfServiceConnectorLimit
            ? fullPricingPerConnector
            : halfPricingPerConnector) *
            connectors,
```

travjenkins: I would make them more like `chargedConnectorLimit` and `connectorLimit`.

I would not use math to calculate 'em and just keep them as 6 and 12 as they are actually just arbitrary numbers.

Brenosalv:
Since the new condition is $50 per connector when it's 7 to 12 connectors, isn't it expected to be $50 x 7 = $350?

Issue #454, opened 2024-08-15
Mobile TOC Hidden on Initial Load


SourabhEstuary:
The table of contents should initially be hidden on mobile devices, only expanding when the user chooses to open it.

Brenosalv:
I also noticed that the table of contents are overflowing the screen when the number of headings are high. We can add a scroll in this case.

SourabhEstuary:
scroll option can be best way to make TOC better on both mobile and desktop devices.

travjenkins:
This is live - closing.

PR #456, opened 2024-08-15
Collapse blog post table of contents on mobile

Brenosalv:
#454 

## Changes

-   Collapse table of contents on mobile devices.
-   Add scroll to the table of contents for when it's big enough.

### DIFF
--- a/src/components/BlogPostToc.tsx
+++ b/src/components/BlogPostToc.tsx
@@ -1,5 +1,5 @@
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
-import { Typography } from '@mui/material';
+import { Typography, useMediaQuery } from '@mui/material';
 import Accordion from '@mui/material/Accordion';
 import AccordionDetails from '@mui/material/AccordionDetails';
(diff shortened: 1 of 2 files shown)


Brenosalv:
Added scroll for too tall table of contents section:
image

src/components/BlogPostToc.tsx
```diff
@@ -107,6 +107,7 @@ export const RenderToc = ({ items }: { items: TocItem[] }) => {
         null
     );
     const timeoutRef = React.useRef<NodeJS.Timeout | null>(null);
+    const isMobile = useMediaQuery('(max-width:768px)');
```
travjenkins, line 110: I'm not on my work machine... is this how mobile is normally checked? Just asking as I know there have been a few ways to do it and sometimes they cause issues with rendering.

Brenosalv: Yes, we are always doing like this and I tested - the TOC is collapsed by default on mobile. Also, this hook provides a dynamic value.

travjenkins (approved):
lgtm

Issue #472, opened 2024-08-30
Homepage h1 text with inconsistent line-height


Brenosalv:
**Bug**
Inconsistent line-height from the h1 text on homepage.

**Expected**

- [x] Replace or remove inconsistent h1's line-height;

**Screenshots**
<img width="444" alt="image" src="https://github.com/user-attachments/assets/ff17683b-ce30-4ab9-afeb-48f82518d061">

PR #473, opened 2024-08-30
Remove individual line-height from h1 styles

Brenosalv:
#472 

## Changes

-   Fix line-height of the h1 from the homepage section one.
-   This issue seems to happen only in production.
-   It was being set twice in the code:
image

## Tests / Screenshots

-   Tested locally:
image


travjenkins (approved):
lgtm

PR #475, opened 2024-09-02
Warning about conflicts with  css order of importing 

Brenosalv:
## Changes

-   Ignore the warning since CSS Modules are scoped to a component and the order of importing does not matter. 

## Tests / Screenshots

-   Tested locally and the warnings are not displayed.

### DIFF
--- a/gatsby-node.ts
+++ b/gatsby-node.ts
@@ -2,6 +2,7 @@
 import pg from 'pg';
 import { GatsbyNode } from 'gatsby';
 import { createRemoteFileNode } from 'gatsby-source-filesystem';
+import FilterWarningsPlugin from 'webpack-filter-warnings-plugin';
 import { SUPABASE_CONNECTION_STRING } from './config';
 import { normalizeConnector } from './src/utils';
 
@@ -443,5 +444,16 @@ export const createResolvers: GatsbyNode['createResolvers'] = async ({
                 },
             },
         },
+    });
+};
+
+exports.onCreateWebpackConfig = ({ actions }) => {
+    actions.setWebpackConfig({
+        plugins: [
+            new FilterWarningsPlugin({
+                exclude:
+                    /mini-css-extract-plugin[^]*Conflicting order. Following module has been added:/,
+            }),
+        ],
     });
 };
--- a/gatsby-node.ts
+++ b/gatsby-node.ts
@@ -447,7 +447,7 @@ export const createResolvers: GatsbyNode['createResolvers'] = async ({
 };
 
 export const onCreateWebpackConfig = ({ stage, actions, getConfig }) => {
-    if (stage === 'develop') {
+    if (['build-javascript', 'develop'].includes(stage)) {
         const config = getConfig();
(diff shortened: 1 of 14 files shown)


gatsby-node.ts
```diff
@@ -445,3 +446,14 @@ export const createResolvers: GatsbyNode['createResolvers'] = async ({
         },
     });
 };
+
+export const onCreateWebpackConfig = ({ actions }) => {
```
travjenkins, line 450: Can you mark this with a TODO and a link to the permanent solution with `ignoreOrder`

Brenosalv: Could you share the link? I thought it could be the permanent solution.

travjenkins: Not sure if Gatsby is using a version of WebPack that would allow us to use this - but this is the true permanent solution.

https://webpack.js.org/plugins/mini-css-extract-plugin/#ignoreorder 
https://www.gatsbyjs.com/docs/how-to/custom-configuration/add-custom-webpack-config/

Brenosalv: Added ignoreOrder as main solution while keeping current solution as a backup.

Brenosalv: Added `ignoreOrder` as main solution while keeping current solution as a backup.

Brenosalv:
Fixing another warning:
image

It seems like this happens when we set `align-items: start or end` instead of `align-items: flex-start or flex-end`. I added `flex` and the warning disappeared.

gatsby-node.ts
```diff
@@ -445,3 +445,16 @@ export const createResolvers: GatsbyNode['createResolvers'] = async ({
         },
     });
 };
+
+export const onCreateWebpackConfig = ({ stage, actions, getConfig }) => {
+    if (stage === 'develop') {
```
travjenkins, line 450: I think we should probably keep this on for `build-javascript` as well.

I know the warning does not show up in prod but probably good to keep these configs in sync.

src/components/Footer/styles.module.less
```diff
@@ -0,0 +1,159 @@
+@import '../../globalStyles/sections.module.less';
+
+footer {
```
travjenkins, line 3: How does css modules handle when you are styling an element in the root of the file?

Brenosalv: It handles the same way it's done with pure css. Just add the styles to the element:
image

travjenkins: Thanks.

So this would end up potentially causing issues with ordering... since there is no unique wrapper applied.

I don't think we have anything else styling `footer` broadly... I think we should still go ahead and convert this to a class to remove any change of issues in the future.

Brenosalv: Makes sense, it would avoid a possible conflict in the future. Even considering there is no sense on adding another footer element, it's still possible.

Brenosalv: image

travjenkins (approved):
lgtm - pulled locally and ran a build and worked fine

PR #480, opened 2024-09-05
Split global CSS code into smaller chunks

Brenosalv:
It continues to solve the issue #429.

## Changes

-   Continue the code split of the global css file `style.less`.


src/pages/about/index.tsx
travjenkins, line 1: In the future we should break this file up a bit. Each section could get a stand alone component. Also, I think this is a good place to move some styling into nested selectors in the parents. Like `sectionOneH1` could easily just be within the `sectionOne` class

src/pages/solutions/index.tsx
```diff
@@ -1,97 +1,115 @@
...
 
 const SolutionsPage = () => {
     return (
         <Layout>
-            <div className="product-flow-section-one-background-image-wrapper">
-                <div className="product-flow-section-one">
-                    <div className="product-flow-section-one-left">
-                        <h1 className="product-flow-section-one-h1">
+            <div className={sectionOneBackgroundImageWrapper}>
+                <div className={sectionOne}>
```
travjenkins, line 36: Can you move the `H1` and `SubText` styling to being nested under here since they are only applied to single tags.

src/pages/solutions/index.tsx
```diff
@@ -1,97 +1,115 @@
...
                 </div>
             </div>
-            <div className="solutions-banner-wrapper">
+            <div className={bannerWrapper}>
                 <SolutionsBanner />
             </div>
-            <div className="solutions-section-wrapper">
-                <div className="solutions-section-text-wrapper">
-                    <h2 className="product-flow-section-two-topic-header">
+            <div className={sectionWrapper}>
```
travjenkins, line 58: I think these can be moved up to `sectionWrapper`

`sectionTwoTopicHeader` -> `h2`
`subtext` -> `p`

src/pages/solutions/styles.module.less
```diff
@@ -0,0 +1,165 @@
...
+  color: #47506d;
+  font-weight: 700;
+  margin-top: 40px;
+  line-height: 30px;
+
+  @media (max-width: 810px) {
+    font-size: var(--fontSize-0);
+    line-height: 22px;
+  }
+}
```
travjenkins, line 101: If we apply `subtext` to all nested `p` then these can be removed:

`color`, `line-height`, and the `media query`

travjenkins (approved):
lgtm

Issue #482, opened 2024-09-09
Add link to status page 


travjenkins:
## Description

Want to link to our status page from marketing

## Work required

-   [x] Add link to footer of page
- [x] Add link to header under `Resources`

## Figma / Designs

![image](https://github.com/user-attachments/assets/36a8c717-51de-42f7-a9ad-6633fe58c142)

PR #483, opened 2024-09-09
Add link to status

travjenkins:
https://github.com/estuary/marketing-site/issues/482

### 482
- Added link to footer
- Added link and icon to the header

## Tests / Screenshots

image

### DIFF
--- a/src/components/HeaderNavbar/Resources/items.tsx
+++ b/src/components/HeaderNavbar/Resources/items.tsx
@@ -28,6 +28,8 @@ export const read = [
             <StaticImage
                 src="../../../svgs/product-flow-live-reporting.svg"
                 alt="Status"
+                height={18}
+                width={18}
             />
         ),
     },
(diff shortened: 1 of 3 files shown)


src/components/HeaderNavbar/Resources/items.tsx
```diff
@@ -20,6 +20,19 @@ export const read = [
...
+    {
+        name: 'Status',
+        to: statusPageUrl,
+        hasChevronIcon: true,
+        Image: () => (
+            <StaticImage
+                src="../../../svgs/product-flow-live-reporting.svg"
+                alt="Status"
+                height={18}
+                width={18}
```
Brenosalv, line 32: Since the other icons don't have width and height set, should we set these values to this one?

Brenosalv (approved):
Lgtm - Left just one comment.

src/components/HeaderNavbar/Resources/items.tsx
travjenkins: Yeah... if we don't the icon ends up looking too big. I am just using an icon we already have and will ask Polina to come up with one that will work here better as a fast follow.

Issue #479, opened 2024-09-05
Create a dedicate page for all authors


Brenosalv:
Create author page: https://www.figma.com/design/qhXFvOJPqLgGsr9lsAPYbn/Estuary%3A-Product-Page?node-id=2451-6411&t=Cm4oSOapwI5WH94S-0

PR #492, opened 2024-09-16
Create author page

Brenosalv:
#479 

## Changes

-   Create a page dedicated to the author.

### DIFF
--- a/gatsby-node.ts
+++ b/gatsby-node.ts
@@ -363,7 +363,7 @@ export const createPages: GatsbyNode['createPages'] = async ({
             allStrapiAuthor {
                 nodes {
                     id
-                    name: Name
+                    slug: Slug
                 }
             }
         }
@@ -372,7 +372,7 @@ export const createPages: GatsbyNode['createPages'] = async ({
     if (authors.data?.allStrapiAuthor.nodes) {
         for (const author of authors.data.allStrapiAuthor.nodes) {
             createPage({
-                path: author.name.replace(' ', '-').toLowerCase(),
+                path: `/author/${author.slug.toLowerCase()}`,
                 component: authorComponent,
                 context: {
                     id: author.id,
--- a/gatsby-node.ts
+++ b/gatsby-node.ts
@@ -26,6 +26,8 @@ const caseStudyTemplate = path.resolve('./src/layouts/CaseStudy/index.tsx');
 const connector = path.resolve('./src/templates/connector/index.tsx');
 const connection = path.resolve('./src/templates/connection.tsx');
 
+const authorComponent = path.resolve('./src/templates/author/index.tsx');
+
 export const createPages: GatsbyNode['createPages'] = async ({
     graphql,
     actions,
@@ -349,6 +351,33 @@ export const createPages: GatsbyNode['createPages'] = async ({
                     });
                 }
             }
+        }
+    }
+
+    const authors = await graphql<{
+        allStrapiAuthor: {
+            nodes: any[];
+        };
+    }>(`
+        {
+            allStrapiAuthor {
+                nodes {
+                    id
+                    name: Name
+                }
+            }
+        }
+    `);
+
+    if (authors.data?.allStrapiAuthor.nodes) {
+        for (const author of authors.data.allStrapiAuthor.nodes) {
+            createPage({
+                path: author.name.replace(' ', '-').toLowerCase(),
+                component: authorComponent,
+                context: {
+                    id: author.id,
+                },
+            });
         }
     }
 };
--- a/gatsby-node.ts
+++ b/gatsby-node.ts
@@ -4,6 +4,7 @@ import { GatsbyNode } from 'gatsby';
 import { createRemoteFileNode } from 'gatsby-source-filesystem';
 import { SUPABASE_CONNECTION_STRING } from './config';
 import { normalizeConnector } from './src/utils';
+import { getAuthorPathBySlug } from './shared';
 
 /**
  * Implement Gatsby's Node APIs in this file.
@@ -372,7 +373,7 @@ export const createPages: GatsbyNode['createPages'] = async ({
     if (authors.data?.allStrapiAuthor.nodes) {
         for (const author of authors.data.allStrapiAuthor.nodes) {
             createPage({
-                path: `/author/${author.slug.toLowerCase()}`,
+                path: getAuthorPathBySlug(author.slug),
                 component: authorComponent,
                 context: {
                     id: author.id,
--- a/src/components/AuthorBlogPostCard/index.tsx
+++ b/src/components/AuthorBlogPostCard/index.tsx
@@ -12,6 +12,7 @@ import {
     articleCardAuthors,
     articleCardFooter,
     dot,
+    authorInfo,
 } from './styles.module.less';
 
 interface AuthorBlogPostCardProps {
@@ -32,6 +33,8 @@ const AuthorBlogPostCard = ({ data: blogPost }: AuthorBlogPostCardProps) => {
         );
 
     const authorName = blogPost?.authors[0]?.name;
+
+    const authorRole = blogPost?.authors[0]?.role;
 
     return (
         <Link to={`/${blogPost?.slug}`} className={container}>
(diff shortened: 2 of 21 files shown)


gatsby-node.ts
```diff
@@ -351,6 +353,33 @@ export const createPages: GatsbyNode['createPages'] = async ({
...
+                    name: Name
+                }
+            }
+        }
+    `);
+
+    if (authors.data?.allStrapiAuthor.nodes) {
+        for (const author of authors.data.allStrapiAuthor.nodes) {
+            createPage({
+                path: author.name.replace(' ', '-').toLowerCase(),
```
travjenkins, line 375: This is wrong. We should be using `slug` for this. 

Also - we should probably make this something like
```
path: `/author/${slug.toLowerCase()}`
```

We should check with @SourabhEstuary on what path is best. As I think we might end up with`.../authors` to show all authors. Then I think it makes sense to make the URL clearly singular with `.../author/...`.

src/components/AuthorBlogPostCard/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+    articleCardFooter,
+    dot,
+    authorInfo,
+} from './styles.module.less';
+
+interface AuthorBlogPostCardProps {
+    data: any;
+}
+
+const AuthorBlogPostCard = ({ data: blogPost }: AuthorBlogPostCardProps) => {
```
travjenkins, line 22: What about instead of having `blogPost?` all over we should check for that right at the start and return null. Maybe throw an error so we could catch it?

src/components/AuthorBlogPostCard/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+    articleDateAndTime,
+    articleCardAuthors,
+    articleCardFooter,
+    dot,
+    authorInfo,
+} from './styles.module.less';
+
+interface AuthorBlogPostCardProps {
+    data: any;
+}
```
travjenkins, line 20: Instead of checking for an empty post could we fix this typing?

src/components/Bio/index.tsx
```diff
@@ -30,7 +31,10 @@ const Bio = ({ authors }: BioProps) => {
                 );
 
                 const combined = (
-                    <div className={avatarWrapper}>
+                    <Link
+                        to={`/${name.replace(' ', '-').toLowerCase()}`}
```
travjenkins, line 35: As called out this link needs to change. Also, we should make a function that generates these so it is always the same.

src/templates/author/SectionOne/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+                                    .gatsbyImageData
+                            }
+                            loading="eager"
+                        />
+                    </div>
+                ) : null}
+                <div
+                    className={clsx(
+                        authorInfo,
+                        bio?.data.bio ? null : centralizeAuthorInfo
```
travjenkins, line 48: We should try to not check this multiple times and just check it and set a class on the parent.

src/templates/author/SectionOne/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+                        className={clsx(
+                            socialIconsWrapper,
+                            bio?.data.bio ? null : centralizeAuthorInfo
+                        )}
+                    >
+                        <SocialLinks
+                            socialLinks={socials}
+                            hasHighlightedIcons
+                        />
+                    </div>
```
travjenkins, line 68: Social Links already have a wrapper element - what is driving having another one here?

src/templates/blog-post/index.tsx
```diff
@@ -397,6 +336,7 @@ const BlogPostTemplate = ({ data: { post } }) => {
                     </div>
                 </section>
                 <section className={popularArticlesWrapper}>
+                    <h2>Popular Articles</h2>
```
travjenkins, line 339: Is it good to have two `h2` tags on the same page?

src/templates/blog-post/index.tsx
```diff
@@ -255,15 +247,15 @@ const BlogPostTemplate = ({ data: { post } }) => {
                         </div> */}
 
                     <div className={aboutAuthor}>
-                        <h3>
+                        <h2>
```
travjenkins, line 250: This is the second `h2` i am referencing down below.

gatsby-node.ts
Brenosalv: I totally forgot the `slug` 🤦🏻‍♂️

Brenosalv: I totally forgot the slug existence 🤦🏻‍♂️

src/components/AuthorBlogPostCard/index.tsx
Brenosalv: Actually blogPost should not be optional. I will remove the question mark.

src/components/AuthorBlogPostCard/index.tsx
Brenosalv: You're right. Done.

src/templates/author/SectionOne/index.tsx
Brenosalv: No need. Removing.

src/templates/author/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+    data: {
+        author: { name, bio },
+    },
+}) => {
+    return (
+        <Seo
+            title={name}
+            description={
+                bio.data.bio
+                    ? bio.data.bio.replace('<p>', '').replace('</p>', '')
```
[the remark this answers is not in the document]
Brenosalv: We can use a regex to remove the HTML tags.

src/templates/blog-post/index.tsx
Brenosalv: Yes, the title of each section is an `h2`. We just can't have two `h1`.

By the way, the text below the `h1` should be a paragraph `p`, not an `h2`. Fixed that.

travjenkins: Oh okay cool - wasn't sure if other headers repeating was bad - but makes sense.

src/templates/author/index.tsx
```diff
@@ -45,7 +52,7 @@ export const Head = ({
             title={name}
             description={
                 bio.data.bio
-                    ? bio.data.bio.replace('<p>', '').replace('</p>', '')
+                    ? bio.data.bio.replace(htmlTagRegex, '')
```
travjenkins, line 55: This is pretty risky. I think we should just grab a library for this to make sure it is as safe as possible.

Something like https://www.npmjs.com/package/string-strip-html seems alright

travjenkins: https://codsen.com/os/string-strip-html/examples#remove-html

I think this is the specific api we'd want

Brenosalv: This one seems to be a good option as well: https://www.npmjs.com/package/html-to-text

Brenosalv: @travjenkins This one seems to be a good option as well: https://www.npmjs.com/package/html-to-text

gatsby-node.ts
SourabhEstuary: I agree that using /authors to show all authors and /author/author-name for individual pages makes the most sense. Let's proceed with that approach.

src/templates/author/index.tsx
Brenosalv: It works with nested elements using the `html-to-text` library.

Tested the following:

```{htmlToText('<p><strong>test_1</strong><div style="width=100%">test_2</div> test_3 <a href="https://example.com">test_4</a> <span>test_5</span></p>', { wordwrap: false })}```

Result: test_1 test_2 test_3 test_4 [https://example.com] test_5

travjenkins: I think that's totally fine.

travjenkins (approved):
lgtm

src/templates/author/index.tsx
Brenosalv: image

PR #497, opened 2024-09-18
Remove Link property from author queries

Brenosalv:
## Changes

-   Remove deprecated Link property from authors queries.


travjenkins (approved):
lgtm

Issue #457, opened 2024-08-15
New ETL Tools comparisons web page


EstuaryRob:
## Description

New Comparison page (basically allow people to compare any ETL tool with each other and add Estuary to it though not in the title.

Dynamic content pages that are crawl-able via google. Firebolt website as example. 

Want to add a new page for all comparisons with the URL etl-tools, and generate comparison pages based on the ELT and ETL comparison guides. See document for more.

## Details Writeup

https://docs.google.com/document/d/1LyJRQQPhdnMRWUoS5hfl7USlWkUzJJVqQCUPwbXUPII/edit?usp=sharing

## Technical work

- [x] Design data format and add to Strapi https://github.com/estuary/strapi-admin/pull/22
- [x] Add content into Strapi
- [x] Build this page a lot like comparison pages to be auto generated with unique URLS
- [x] Provide a way for user to select which items to compare

travjenkins:
We'll want to update the strapi work to include there changes:
![image](https://github.com/user-attachments/assets/be8b06a4-db9e-4129-b9a5-157da66bf1ae)

PR #503, opened 2024-09-23
Add comparison pages

Brenosalv:
#457 

## Changes

-   Create comparison center page;
-   Create comparison XvsY page.


src/pages/etl-tools/index.tsx
```diff
@@ -44,6 +46,11 @@ export default EtlTools;
 
 export const pageQuery = graphql`
     query GetAllComparisons {
+        estuaryVendor: strapiComparison(
+            id: { eq: "d829928c-c473-5421-ac0a-f03c45b14993" }
+        ) {
+            id
+        }
```
Brenosalv, line 53: Fetching only the estuary id seems redundant but if the id changes in strapi, we'll know because the query won't work anymore.

We are using the estuary vendor id for setting the estuary tab as default here:
image

.lighthouserc.js
```diff
@@ -16,7 +16,7 @@ module.exports = {
                 'http://localhost:9000/integrations/index.html',
                 'http://localhost:9000/source/postgres/index.html',
                 'http://localhost:9000/integrations/postgres-to-elasticsearch/index.html',
-                'http://localhost:9000/vs-fivetran/index.html',
+                'http://localhost:9000/etl-tools/estuary-vs-fivetran/index.html',
```
travjenkins, line 19: We'll need redirects for `confluent` as well.

Brenosalv: I added redirects for all the comparison links from the menu. I'm considering all of them are indexed.

src/components/BlogPostToc/index.tsx
```diff
@@ -192,7 +192,11 @@ export const RenderToc = ({ items }: { items: TocItem[] }) => {
                     className={accordionSidePadding}
                     expandIcon={
                         <ExpandMoreIcon
-                            sx={{ color: '#47506d', fontSize: '2rem' }}
+                            sx={{
+                                color: '#47506d',
+                                fontSize: '2rem',
+                                maxWidth: '36px',
```
travjenkins, line 198: Why is this styling not park of the `accordionSidePadding` styling?

src/components/EtlToolsPage/SectionTwo/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+                </h2>
+                <div className={cardsWrapper}>
+                    <Card
+                        title="The Data Engineer's Guide to ELT Alternatives"
+                        buttonHref="/ELT-comparison-guide"
+                    />
+                    <Card
+                        title="The Data Engineer's Guide to ETL Alternatives"
+                        buttonHref="/ETL-alternatives-guide"
+                    />
```
travjenkins, line 22: the `title` is the same on both of these cards... I think the comparison one needs updated.

src/components/EtlToolsPage/shared.ts
travjenkins, line 1: Looks like an empty file - I think we can delete?

src/components/EtlToolsXvsYPage/SectionThree/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <section className={defaultWrapperDark}>
+            <div className={container}>
+                <h2>GETTING STARTED WITH ESTUARY</h2>
+                <ul>
+                    <FeatureCard
+                        icon={
+                            <EstuaryLogo
+                                width={28}
+                                height={28}
+                                color="#5072EB"
```
travjenkins, line 28: We are using `#5072EB` in a lot of places here. Can we get that shared in a single const?

src/components/EtlToolsXvsYPage/SectionTwo/Connectors/index.tsx
travjenkins, line 1: All these stand alone components are basically the same. We need to make a single template that they all pull from.

src/pages/etl-tools/index.tsx
travjenkins: Yeah... I that makes sense. That way it changing could be caught by breaking the build.

travjenkins (changes requested):
The only major change we really need to making all the table row stuff to be shared.

src/components/EtlToolsPage/SectionTwo/index.tsx
Brenosalv: Actually one is for "ELT" and other for "ETL". Don't you think it's better to keep them independent?

src/components/EtlToolsXvsYPage/SectionThree/index.tsx
Brenosalv: Yeah, the same for the 28. I will create two consts out of the component scope.

src/components/EtlToolsXvsYPage/index.tsx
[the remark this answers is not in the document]
Brenosalv: Oh, I wasn't aware of this - thanks for the heads up.

Brenosalv:
Build failed due to Strapi changes. I will re-run the site build once the strapi build finishes.

src/components/EtlToolsPage/SectionTwo/index.tsx
travjenkins: Oh I see... but still shouldn't the `title` match the `href`? 

The first linke os to a "comparison guide" so I would think the title would be "The Data Engineer's Guide to ELT Comparisons"

Brenosalv: Oh yeah, I followed the design and didn't notice this. Thanks.

Brenosalv: Checking the blog post title, the title really is "Alternatives":
image

src/components/BlogPostToc/index.tsx
Brenosalv: No reason, it should be part of the class. Fixed.

Brenosalv:
A placeholder was added to the select on the right as default value:
image

src/components/EtlToolsXvsYPage/SectionTwo/TableRows/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+    rows: Row[];
+    xVendor: ComparisonVendor;
+    yVendor: ComparisonVendor;
+    estuaryVendor: ComparisonVendor | null;
+    dataType: string;
+}
+
+const getNestedProperty = (obj: Record<string, any>, path: string) => {
+    return path.split('.').reduce((acc, key) => acc?.[key], obj);
+};
```
travjenkins, line 26: This is clever. We could have used a JSONPath library but this is totally fine. We will want to keep this in mind if we add more data that the names cannot contain a `.` in them.

src/components/EtlToolsXvsYPage/SectionTwo/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+                }
+            });
+        };
+
+        window.addEventListener('scroll', handleScroll);
+
+        return () => {
+            window.removeEventListener('scroll', handleScroll);
+        };
+    }, []);
```
travjenkins, line 133: Instead of 'scroll` we should be using an `IntersectionObserver` that can tell us when something is "sticky".

https://davidwalsh.name/detect-sticky

travjenkins: I can take care of this quickly

travjenkins: Ugh... never mind... I don't wanna hold this up so we can do this later.

travjenkins (dismissed):
lgtm - just look into that last error popping up in the console

src/components/EtlToolsXvsYPage/SectionTwo/TableRows/index.tsx
Brenosalv: Yeah, I implemented this with scalability in mind, allowing it to handle any object, regardless of the number of nested properties.

travjenkins (approved):
lgtm

travjenkins (approved):
lgtm

Issue #534, opened 2024-10-11
Duplicate page for comparison X vs Y and Y vs X


Brenosalv:
**Bug**
Both X-vs-Y and Y-vs-X pages are being generated with the same content.

**Expected**
- Remove duplicate content by keeping it always X-vs-Y and never Y-vs-X.
- Redirect Y-vs-X pages to X-vs-Y ones since they were already indexed.

PR #535, opened 2024-10-11
Remove duplicate pages for comparison x vs y and y vs x

Brenosalv:
#534 

## Changes

-   Removed duplicate page contents and URLs from comparison pages.
-   Added all the wrong indexed pages to redirect.
-   Now we generate all the pages in ascendent alphabetical order.

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -152,28 +152,28 @@
                 "type": 301
             },
             {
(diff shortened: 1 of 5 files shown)


travjenkins (approved):
lgtm

PR #540, opened 2024-10-16
Add missing ai pipelines field and remove two fields

Brenosalv:
## Changes

-   Add missing `AI_Pipelines` field into comparison table > Use cases, and remove `Data_Science_ML` and `Historial_Analytics` fields as requested by Dani

-  We should make the same changes in Strapi Admin code.


travjenkins (approved):
lgtm

Issue #528, opened 2024-10-08
Add security certs to main page


danthelion:
Similarly as they are on the product page: https://estuary.dev/product/

position: below case studies

PR #541, opened 2024-10-16
Add security certificates section on the homepage

Brenosalv:
#528

## Changes

-   Add security certificates section below case studies section on the homepage.

## Tests / Screenshots
Below the case sturdies section:
image


Brenosalv:
@danthelion Adding the security certificates section on the homepage conflicts the alternate colors. Is it okay for you?

image

danthelion:
@Brenosalv that won't work, let's re-think how we add this again.

src/components/Homepage/AutomateDataops/index.tsx
```diff
@@ -13,7 +13,7 @@ import {
     Wrapper,
 } from './styles';
 
-const SectionSeven = () => {
+const AutomateDataops = () => {
```
travjenkins, line 16: :tada: :tada: :tada: :tada: :tada: :tada: :tada: 

Thank you so much for naming these something that makes sense :D

Brenosalv: Haha actually it made sense to me since the sections have an order. However I noticed it wasn't scalable when I needed to add new sections. 😅

To be honest, I didn't like to name the file with the title of the section because it makes them dependent on each other. If we need to change the text of the title, we also have to change the name of the file. But yeah, I think it's the best approach anyway.

travjenkins: I would recommend instead of naming them too closely to what the message is - name them more oh what the "idea" of the messaging is.

src/components/Homepage/UseTheBestRealTimeCdc/index.tsx
```diff
@@ -14,7 +14,7 @@ import {
     Wrapper,
 } from './styles';
 
-const SectionFour = () => {
+const UseTheBestRealTimeCdc = () => {
```
travjenkins, line 17: Since we are using react we should probably not start a component with `use`.

src/components/Homepage/AutomateDataops/index.tsx
Brenosalv: It also works well but the reason to name them exactly was to make it easier to locate the section with a `command + F`.

src/components/Homepage/Do4xMoreWith25xLess/index.tsx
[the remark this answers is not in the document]
Brenosalv: Yes, I think it will work well like this.

travjenkins (approved):
lgtm

Issue #513, opened 2024-10-01
Deployment Options page


danthelion:
## Description

New page for deployment options
[Estuary Flow Deployment Modes](https://docs.google.com/document/d/1zp6weHNsMqUVNey9Kz02CmPabSCT3befUX7H76g39tc/edit)

## Work required

-   [x] Create deployment options page.

## Figma / Designs

https://www.figma.com/design/qhXFvOJPqLgGsr9lsAPYbn/Estuary%3A-Product-Page?node-id=8661-24807&node-type=frame&t=xHXFeBBVtswEVbcu-0

Brenosalv:
@danthelion Is it ready for development or still need design? Also, I don't have access to the doc from the description. I requested access.

PR #543, opened 2024-10-17
Create deployment options page

Brenosalv:
#513 

## Changes

-   Create deployment options page

The path/slug is `/deployment-options`.


src/components/DeploymentOptionsPage/shared.ts
```diff
@@ -0,0 +1,23 @@
+export const features = {
```
travjenkins, line 1: Instead of having a stand alone list in the type `DeploymentOption` we should just reference this. That way it will always be in sync.

```
export type DeploymentOption = keyof typeof features;
```

src/pages/deployment-options/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+                }
+            }
+        }
+    `);
+
+    return (
+        <Seo
+            title="Estuary Flow Deployment Options: Public, Private & BYOC"
+            description="Estuary Flow offers Public, Private, and BYOC deployment options for secure, scalable, and customizable data integration, perfect for any business needs."
+            image={metaImg.childImageSharp.gatsbyImageData.images.fallback.src}
```
travjenkins, line 42: Looks good!
image

travjenkins (commented):
One small change to typing

src/components/DeploymentOptionsPage/shared.ts
Brenosalv: Makes sense, thanks.

Brenosalv:
I created a container component and applied it to all the sections. I still need to do this for sections of the other pages.

travjenkins (approved):
lgtm

Issue #544, opened 2024-10-17
Contact Us CTA Background Display Issue (ETL Tools Comparison Page)


SourabhEstuary:
**Bug**
Background of the "Contact Us" CTA at the bottom of the ETL tools comparison page appears white.

**Screenshots**
![image](https://github.com/user-attachments/assets/c0e6e596-2bec-4b64-92ad-24f8b33e038d)

PR #545, opened 2024-10-17
Fix banner background color from etl tools XvsY page

Brenosalv:
#544 

## Changes

-   Fix banners background image from ETL Tools XvsY page 

## Tests / Screenshots

-  etl-tools page X vs Y:
image

image


travjenkins (approved):
lgtm

Issue #542, opened 2024-10-17
Create a component for the hero section


Brenosalv:
## Description

The hero sections of various pages have the same structure. Create a component for it.

## Work required

-   [ ] Create component with title, description, buttons and image;
-   [ ] Add feature tags below buttons as optional prop;
-   [ ] Replace hero sections with this components.
-   [ ] Create a component for the container `div` with the css that is common to all the containers (or use CSS Grid):
```
.globalMaxWidth;
display: flex;
gap: 60px;
align-items: center;
justify-content: space-between;
```

## Designs

Pricing page:
<img width="1447" alt="image" src="https://github.com/user-attachments/assets/49fab5ba-4d33-4958-a5ac-b1d8248c7f63">

Product page:
<img width="1441" alt="image" src="https://github.com/user-attachments/assets/88994cec-3dee-4653-a7b6-56734002dfe1">

Deployment options page:
<img width="939" alt="image" src="https://github.com/user-attachments/assets/e3d088c5-5c17-4e83-b5b2-7fe901859739">

PR #549, opened 2024-10-28
Create components for the page sections

Brenosalv:
#542 

## Changes

It's a refactor to reuse the container div across various page sections.

- Reuse the Container component in most sections.
- Create a `HeroSectionDetails` component including the title, description and buttons of the hero sections. use the homepage left column of the hero section as reference.


src/components/EtlToolsXvsYPage/Hero/index.tsx
```diff
@@ -21,32 +21,41 @@ interface SectionOneProps {
...
-                        <ChevronRightIcon color="#5072EB" />
-                    </NewTabLink>
-                </div>
+            <Container>
+                <HeroSectionDetails
+                    title={
+                        <>
+                            <span>{xVendor.name.toUpperCase()}</span>{' '}
+                            <span>VS</span>{' '}
+                            <span>{yVendor.name.toUpperCase()}</span>
```
travjenkins, line 33: What is driving changing this to uppercase now?

src/components/HeroSectionDetails/index.tsx
travjenkins, line 1: One idea of a future improvement (if we ever need it)

Is instead of having `hasCtaButtons` make that be an optional param with the name `ctaButtons` and then default it to the buttons you normally want there.

That way we can pass in any buttons we need or set to `null` if we want no buttons.

src/components/EtlToolsXvsYPage/Hero/index.tsx
Brenosalv: I applied uppercase to the entire `h1` text and forgot to remove this. Thanks.

src/components/HeroSectionDetails/index.tsx
Brenosalv: I considered only these two buttons because they were the most used but I like your suggestion since we might have different buttons eventually in the future. I will have this in mind.

src/templates/connector/Hero/Content/styles.ts
```diff
@@ -1,56 +1,43 @@
...
+    margin-bottom: 32px;
     margin: 0;
-    color: #47506d;
-    text-transform: uppercase;
-
-    @media (min-width: 1280px) {
-        font-size: 24px;
-        line-height: 29px;
-    }
-`;
```
Brenosalv, line 38: > Since we are getting stuff to be shared... this one might need some tweaks to make look better
> 
> I think we need some space between the icon line and the title
> 
> Also - I think we might want to add support for like a "standard" and a "small" size for the title text? Just an idea on how to handle this.
> 
> image

That's the reason. There was a `margin: 0;` after the `margin-bottom: 32px;`.

travjenkins (approved):
lgtm - there are subtle changes to the heros but they all look in sync and I think they all look good.

Issue #539, opened 2024-10-16
Implement Author Schema Markup for Author Pages


SourabhEstuary:
### Please add the following structured data (Schema.org Person type) for our dedicated author pages.

Author Schema - Dave
Page URL: https://estuary.dev/author/dave/

{
  "@context": "https://schema.org",
  "@type": "Person",
  "name": "David Yaffe",
  "url": "https://estuary.dev/author/dave/",
  "jobTitle": "Co-founder and CEO",
  "description": "David Yaffe is the co-founder and CEO of Estuary. He was previously the COO of LiveRamp and co-founder/CEO of Arbor, which was acquired by LiveRamp in 2016. He has a strong background in product management, leading products at Doubleclick Bid Manager and Invite Media.",
  "worksFor": {
    "@type": "Organization",
    "name": "Estuary",
    "url": "https://estuary.dev/"
  },
  "sameAs": [
    "https://www.linkedin.com/in/davidyaffe/",
    "https://www.crunchbase.com/person/david-yaffe",
    "https://twitter.com/dyaffe"
  ],
  "image": "https://estuary.dev/static/fa6288a046089a71163e9e099b7b91a9/c95dc/T0120_UQAX_7_X_U011_VJAMPN_2_e8363e901cd3_512_e6ae747354.webp"
}

------------------------------------------------
Author Schema - Dani
Page URL: https://estuary.dev/author/daniel/

{
  "@context": "https://schema.org",
  "@type": "Person",
  "name": "Dani Pálma",
  "url": "https://estuary.dev/author/daniel/",
  "jobTitle": "Data Professional",
  "description": "Dani is a data professional with expertise in data engineering and real-time platforms. At Estuary, Dani promotes cutting-edge streaming solutions, bridging technical innovation with developer adoption.",
  "worksFor": {
    "@type": "Organization",
    "name": "Estuary",
    "url": "https://estuary.dev/"
  },
  "sameAs": ["https://www.linkedin.com/in/daniel/"],
  "image": "https://estuary.dev/static/364e60c786d179ec822141bba97b5123/c95dc/danippfp_1ba13fb2f9.webp"
}

---------------------------------------------
Author Schema - Rob
Page URL: https://estuary.dev/author/rob/

{
  "@context": "https://schema.org",
  "@type": "Person",
  "name": "Rob Meyer",
  "url": "https://estuary.dev/author/rob/",
  "jobTitle": "Marketing and Product Marketing Professional",
[...]

PR #550, opened 2024-10-29
Adding more detail about authors for SEO

travjenkins:
## Changes

https://github.com/estuary/marketing-site/issues/539
- Made a reusable SEO JSON generator so Blog and Author page can share logic
- Minor tweaks to queries

### Blog post now includes full author details
image

### Author page shows the same deep details
image

### DIFF
--- a/shared.ts
+++ b/shared.ts
@@ -44,8 +44,8 @@ export const getAuthorSeoJson = (author: Author, siteUrl: string) => {
 
     const sameAs: string[] = [];
 
-    if (author.socials.linkedIn) {
-        sameAs.push(author.socials.linkedIn);
+    if (author.socials.linked_in) {
+        sameAs.push(author.socials.linked_in);
     }
 
     if (author.socials.twitter) {
--- a/shared.ts
+++ b/shared.ts
@@ -44,8 +44,8 @@ export const getAuthorSeoJson = (author: Author, siteUrl: string) => {
 
     const sameAs: string[] = [];
 
-    if (author.socials.linked_in) {
-        sameAs.push(author.socials.linked_in);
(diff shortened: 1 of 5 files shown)


shared.ts
```diff
@@ -35,6 +36,48 @@ export const estuaryAddress = {
 export const getAuthorPathBySlug = (slug: string) =>
     `/author/${slug.toLowerCase()}`;
 
+export const getAuthorSeoJson = (author: any, siteUrl: string) => {
```
Brenosalv, line 39: We have an author type available in /src/templates/author/shared.ts.

src/templates/author/index.tsx
```diff
@@ -43,23 +43,33 @@ export default AuthorPage;
...
 export const Head = ({
     data: {
-        author: { name, bio },
+        author,
+        site: {
+            siteMetadata: { siteUrl },
+        },
     },
 }) => {
+    console.log('author', author);
```
Brenosalv, line 52: I think we can remove this console.

Brenosalv (dismissed):
Lgtm - just left two comments for quick fixes.

shared.ts
travjenkins: Oh sweet - wanted sure about that. I'll add it

Brenosalv:
This build is failing because we have blog post without an author. We just need to set the author to this published blog post: https://strapi.estuary.dev/admin/content-manager/collectionType/api::blog-post.blog-post?page=1&pageSize=10&sort=createdAt:DESC&filters[$and][0][authors][Name][$null]=true

Brenosalv (approved):
Lgtm

Issue #525, opened 2024-10-08
New feature banner on main page


danthelion:
Add new feature release banner to main page based on mockup by Polina. 

The banner should be clickable and be configurable where it redirects. The default content should be for the private deployment feature release and redirect should be to the new deployment modes page.

Brenosalv:
I will leave to do this once we have the deployment options page done.

danthelion:
re-usable component for any marketing site page (add to blog page as well and main to start with)

PR #553, opened 2024-10-31
Add a release banner on top of the homepage

Brenosalv:
#525 

## Changes

-   Create a component for the banner and use it on the Layout component;
-   For now, use it only on top of the homepage;
-   Clicking the banner opens a new tab to the deployment options page.

## Tests / Screenshots

-   Desktop:
image

-   Mobile:
image


src/components/Layout/ReleaseBanner/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+            <div>
+                <a href={href} target="_blank" rel="noreferrer">
+                    <span>A Step Ahead</span>
+                    <p>
+                        Explore Estuary&apos;s powerful compliance solutions
+                        &#45; Streamline your security and scale with
+                        confidence!
+                    </p>
+                    <ArrowRight color="#fff" />
+                </a>
```
travjenkins, line 32: We should either pass in everything or nothing. 

Since the plan it for Dani to submit PRs I think we should just make this a stand alone component that only handles taking in `dark` or `light` mode.

Then later on if we need we can make this a base component that takes in an href, title, message, and icon.

src/components/Layout/ReleaseBanner/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+            )}
+        >
+            <div>
+                <a href={href} target="_blank" rel="noreferrer">
+                    <span>A Step Ahead</span>
+                    <p>
+                        Explore Estuary&apos;s powerful compliance solutions
+                        &#45; Streamline your security and scale with
+                        confidence!
+                    </p>
```
Brenosalv, line 30: I'm thinking about passing a prop for this text so it becomes customizable. Not sure if we'll always use this same text for new feature releases.

src/components/Layout/ReleaseBanner/index.tsx
Brenosalv: Okay, so we just need the `href` to be the same for all.

src/components/Layout/ReleaseBanner/index.tsx
Brenosalv: Okay, your first comment answers it.

src/components/Layout/ReleaseBanner/index.tsx
Brenosalv: If in the future we need it to be customizable, we add props to it as you said.

travjenkins (dismissed):
lgtm

src/components/Layout/ReleaseBanner/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+const ReleaseBanner = ({ theme = 'dark' }) => {
+    return (
+        <div
+            className={clsx(
+                container,
+                theme === 'dark' ? darkBackgroundColor : lightBackgroundColor
+            )}
+        >
+            <div>
+                <a href="/deployment-options" target="_blank" rel="noreferrer">
```
travjenkins, line 23: doubt it matters... but should be end this `href` with `/`?

Brenosalv: I wondered the same thing but when we click on the banner, the slash is added automatically. Also, none of the other href have this slash on the end. Anyway, I can ask Sourabh.

travjenkins (approved):
lgtm

Issue #432, opened 2024-08-06
Partnerships Page


altitude-lucas:
General Idea: describes who we are system integration partnering with. Also, CTA to connect with said partners.

danthelion:
mockup: partner page: https://www.figma.com/design/qhXFvOJPqLgGsr9lsAPYbn/Estuary%3A-Product-Page?node-id=7259-24602&node-type=frame&t=1chxOOAqQMqpmoeD-0

PR #560, opened 2024-11-06
Partners page

Brenosalv:
#432

## Changes

-   Create Partners page.


src/components/Container/index.tsx
```diff
@@ -1,27 +1,30 @@
...
 import clsx from 'clsx';
-import { container, vertical, reverse } from './styles.module.less';
+import { container, vertical, reverse, lightText } from './styles.module.less';
 
 interface ContainerProps {
     children: ReactNode;
     className?: string;
     isVertical?: boolean;
     isReverseColumnOnMobile?: boolean;
+    isDarkTheme?: boolean;
```
travjenkins, line 10: Pretty soon I think we need to make a "default" component property interface.

This can include `children`, `className`, `isDarkTheme`, and maybe `isVertical`... since we are constantly using these properties we don't need to keep recreating them over and over.

src/components/PartnersPage/Benefits/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+                        quality={100}
+                        placeholder="blurred"
+                    />
+                    <div className={cardList}>
+                        <Card
+                            icon={<TopPlatformIcon color="#5072EB" />}
+                            title="Co-marketing initiatives"
+                            description="Publish content together with Estuary and cross-market new initiatives and product launches for increased reach."
+                            isDarkTheme
+                        />
```
travjenkins, line 30: As best we can.... for stuff like `description` it is probably better to just use `children` so you aren't passing another prop and also makes the JSX look a bit more like "normal" html.

src/components/PartnersPage/ContactUs/styles.module.less
```diff
@@ -0,0 +1,64 @@
...
+
+.section {
+  display: grid;
+  grid-template-columns: 1fr 1fr;
+  align-items: center;
+  justify-content: center;
+  background-color: #FDFDFE;
+
+  > :nth-child(1),
+  > :nth-child(2) {
```
travjenkins, line 11: Can't this just be `div` and not use an `nth`. I also assume these selectors are more expensive... but that might not be true.

src/components/PartnersPage/JoinPartnerProgram/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+    },
+    {
+        icon: <TopPlatformIcon color="#5072EB" width={40} height={40} />,
+        name: 'Azure Storage',
+    },
+    {
+        icon: <TopPlatformIcon color="#5072EB" width={40} height={40} />,
+        name: 'Azure Storage',
+    },
+];
```
travjenkins, line 42: This is not always true so just mentioning here as this is an example of when it is true.

If you are storing off an array and it contains a bunch of components that all get the same properties sent to them. I think it is usually best to just have the `icon` be the actual `component` and then pass in the props in a single spot.

```
const partners: Partner[] = [
    // Replace this with real logos
    {
        Icon: TopPlatformIcon,
        name: 'Azure Storage',
    },...
```

```
for each partner ({Icon, name})
<Icon color="#5072EB" width={40} height={40} />
...
```

Brenosalv: Actually these icons are just temporary and will be replaced by the real logos, so I didn't care about code repetition. I added a comment above the array but was not clear about this, sorry. When replacing these icons with real logo icons, I will create a separate component.

travjenkins (dismissed):
lgtm - the only callouts are pretty large and systemic potential changes to the codebase

src/components/Container/index.tsx
Brenosalv: I created a new issue for this.

src/components/PartnersPage/ContactUs/styles.module.less
Brenosalv: It's not expensive because we are specifying the children 1 and 2 and we have few children. Also, if we use `div`, it would apply the styles to children of children and we want it to be applied only to the first two divs.

src/components/DeploymentOptionsPage/Hero/index.tsx
```diff
@@ -29,19 +27,7 @@ const Hero = () => {
...
-                                href={dashboardRegisterUrl}
-                                target="_blank"
-                            >
-                                Build a Pipeline
-                            </OutboundLinkFilled>
-                            <LinkOutlined href="/contact-us/" theme="dark">
-                                Contact Us
-                            </LinkOutlined>
-                        </>
-                    }
```
Brenosalv: I just forgot to create a component for that when converting the prop from `hasCtaButtons` to `ctaButtons`. Thanks for doing that.

travjenkins (approved):
Lgtm

PR #568, opened 2024-11-15
Blog post images should lazy load

Brenosalv:
## Changes

-   Set loading attribute of the images to lazy in the patch `@draftbox-co+gatsby-rehype-inline-images+1.0.2.patch`.
-   Remove loading eager from imgSharp.

## Tests / Screenshots

image

Prod env:
image

Dev env:
image

### DIFF
--- a/plugins/estuary-rehype-transformers/src/ImgSharp.jsx
+++ b/plugins/estuary-rehype-transformers/src/ImgSharp.jsx
@@ -1,6 +1,6 @@
 import { GatsbyImage } from 'gatsby-plugin-image';
 
-const ImgSharpInline = ({ className, imgdata, alt, loading }) => {
+const ImgSharpInline = ({ className, imgdata, alt }) => {
     const parsed = JSON.parse(imgdata);
     return (
         <GatsbyImage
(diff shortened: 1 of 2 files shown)


plugins/estuary-rehype-transformers/src/ImgSharp.jsx
```diff
@@ -1,15 +1,13 @@
...
+const ImgSharpInline = ({ className, imgdata, alt, loading }) => {
     const parsed = JSON.parse(imgdata);
     return (
         <GatsbyImage
             className={className}
             style={{ margin: '0 auto' }}
             image={parsed}
             alt={alt || 'Blog Post Image'}
-            loading="eager"
-            placeholder="none"
```
Brenosalv, line 12: This prop does not exist.

Brenosalv:
The images are not showing up due to cache?
image

Brenosalv:
> Forgot that I tried this once and ran into the same issue. I did not look into it at all though - so the fix is possibly simple.

To be honest, I’m not sure how to fix this yet. The lazy loading just works when we refresh the page.

PR #581, opened 2024-11-26
Update logo image URL

Brenosalv:
#570 

## Changes

-   Rename estuary logo.

## Tests / Screenshots

image


Brenosalv:
@travjenkins Not sure if we should and can remove the hash from the end of the logo's name `estuary-430cce9313e0df82d11e40d8285f38b4.png`, but it's much better than `combination-mark__multi-blue-430cce9313e0df82d11e40d8285f38b4.png`.

travjenkins (approved):
lgtm

Brenosalv:
> I don't think we need to remove the hash right now.

Okay, so this is the best we can do. Please feel free to merge.

Issue #476, opened 2024-09-02
Organize website colors into variables


Brenosalv:
## Description

Currently we do not have the right colors from Figma as being the variables in the global css code.

## Work required

-   [ ] Check the colors in Figma and add them to global variables for better maintainability and consistency across the website.

## Figma / Designs

https://www.figma.com/design/qhXFvOJPqLgGsr9lsAPYbn/Estuary%3A-Product-Page?node-id=1526-12433&t=MNnlgB4wMdOuEbuR-0

<img width="343" alt="image" src="https://github.com/user-attachments/assets/e02aed7c-a096-4828-9fcb-a05da58422e2">

PR #591, opened 2024-12-09
Organize website colors into variables

Brenosalv:
#476

## Changes

-  Define variables for each color used in the code based on the Figma design. Extract the color values from Figma and assign them to named variables to ensure consistency and maintainability throughout the codebase.


Brenosalv:
Some colors are used only in specific cases, so I didn’t create variables for them. I added variables only for the most commonly used colors.

travjenkins (approved):
lgtm - pulled the branch and searched for a few colors. 

I think we will want to eventually remove a lot of the uses of variable colors that are one offs (or nearly one offs) with just changing the opacity on the end and force ourselves to only have between 3 and 5 versions of these. That way we can store them all in variables.

Issue #351, opened 2024-06-12
New Customers Page


altitude-lucas:
General idea : getting quotes from customers, champions of the product.

danthelion:
should be a new page to hold all customer caste studies, similarly to: https://www.striim.com/customers

danthelion:
design: https://www.figma.com/design/qhXFvOJPqLgGsr9lsAPYbn/Estuary%3A-Product-Page?node-id=10012-23707&t=Yb6oEEZ4MwauW4wG-1

PR #594, opened 2024-12-12
Create customers page

Brenosalv:
#351

## Changes

-   Create customers page.


src/components/CustomersPage/CaseStudies/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+                                key={caseStudy.id}
+                                data={caseStudy}
+                                footerTag="Case study"
+                            />
+                        ))}
+                </Grid>
+                {visiblePosts < caseStudiesWithPrefixedSlugs.length ? (
+                    <ButtonFilled onClick={handleShowMore}>
+                        Show more
+                    </ButtonFilled>
```
Brenosalv, line 67: I’m considering using a ‘Show more’ button, similar to what we have on the author page, since there are currently only a few case studies. Implementing pagination with multiple pages at this stage would unnecessarily duplicate content across sections. Once we have a larger number of case studies, I’d consider adding proper pagination with ‘Previous’ and ‘Next’ buttons, like we do on the blog page. If you have any objection to this, please let me know.

src/components/CustomersPage/Hero/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+                    {...commonProps}
+                />
+            </li>
+            <li key="revunit">
+                <StaticImage
+                    src="../../../images/customer-logos/revunit.png"
+                    alt="Revunit logo"
+                    {...commonProps}
+                />
+            </li>
```
Brenosalv, line 110: It's repeating because, as we know, the StaticImage does not accept dynamic values. Creating a component would pass the src as prop and it would not work - so just explaining.

src/components/Grid/Card/index.tsx
Brenosalv, line 1: I reused the card component from the author page creating a separate component for it.

src/components/Grid/index.tsx
Brenosalv, line 1: A Grid component was created to be reused in the author, blog and customers pages.

src/components/SeeHowSection/index.tsx
Brenosalv, line 1: A SeeHowSection was created to reuse in the homepage, integration page and customers page with different buttons.

src/components/Grid/index.tsx
Brenosalv: We are now reusing the grid component in the popular articles section from the blog post page as well.

src/components/CustomersPage/CaseStudies/index.tsx
travjenkins: I agree we should not worry about this right now as we don't have too many. We could also just not use pagination and list them all the time like Connectors page does.

travjenkins (dismissed):
lgtm

travjenkins (approved):
lgtm

Issue #578, opened 2024-11-25
Optimize hook imports for Tree Shaking benefits


Brenosalv:
## Description

Maybe we could get some tree shaking benefits if we switched from using `React.use...` to just importing the specific hooks we need.

## Work required

-   [x] Check if we really need to remove the `React.`;
-   [x] Maybe we should remove all the `React.` and import directly from the react lib.

PR #607, opened 2024-12-19
Optimize hook imports for tree shaking benefits and remove unnecessary React imports

Brenosalv:
#578

## Changes

-   Remove React. to access react properties and use them directly from the react import;
-   Remove all the `import React from 'react'` and add a plugin to add React to all component instances. The plugin: https://www.gatsbyjs.com/plugins/gatsby-plugin-provide-react/

Obs: Gatsby does not allow removing the `React` import without the plugin, even with a react version 17+ that does not need this import anymore. That's why we are using the plugin.


travjenkins (approved):
lgtm

PR #613, opened 2025-01-02
Link to the AICPA site for the SOC2 logo

travjenkins:
## Changes

- We need to link to this site from the logo so just wrapping the entire image as it is not broken up.

## Tests / Screenshots


Brenosalv:
In terms of UX, it doesn't look clear that there is a link attached to the image. Maybe a button below the current button would look better in terms of UX. Just a suggestion in case this link is very important to be clicked.

Brenosalv (approved):
Lgtm

Issue #434, opened 2024-08-06
Solutions Pages


altitude-lucas:

altitude-lucas:
@EstuaryRob discussing outline with Danny

danthelion:
design: https://www.figma.com/design/qhXFvOJPqLgGsr9lsAPYbn/Estuary%3A-Product-Page?node-id=9835-23602&node-type=section&t=jP9bXwyNPSs6Aa2S-0

Brenosalv:
@danthelion It seems like we have two versions of the solutions page in Figma. Which one would you like me to implement?

danthelion:
@Brenosalv here is the brief for the solutions page: https://docs.google.com/document/d/1ogyP2ubvRazNz_Cv3RT3ZnakNXpyow1L06D7EKjtCOY/edit?tab=t.0#heading=h.pihj74sbtd03

There's only one version, for various pages. Not all have a design yet but I thought we could start implementing them one by one

Brenosalv:
@danthelion Got it. Did you moved this to backlog intentionally?

danthelion:
In total we'll have 3 designs, 1 for each solution category: Use Cases, Industry and Technology

We have 2 deisngs ready, we can use them for Use Case pages and for Industry pages.

All use case pages should use the "Data Movement" design.

All Industry pages should use the "Data warehouse design"

Technology pages design WIP.

+ implement navigation (top-level Solutions text) + adjust responsiveness of logo

navigation panel has 3 columns for each category.

PR #618, opened 2025-01-06
Solutions pages

Brenosalv:
#434 

## Changes

- Create templates for the "Use Case", "Industry" and "Technology" solution pages, totalizing 11 pages;
- Add the page contents from the docs directly to the code.


gatsby-node.ts
```diff
@@ -488,19 +488,51 @@ export const createPages: GatsbyNode['createPages'] = async ({
...
 
+    const useCaseSolutions = allSolutions.filter((solution) =>
+        solution.slug.includes('/use-cases/')
+    );
+    const industrySolutions = allSolutions.filter((solution) =>
+        solution.slug.includes('/industry/')
+    );
+    const technologySolutions = allSolutions.filter((solution) =>
+        solution.slug.includes('/technology/')
+    );
```
Brenosalv, line 499: I'm using the slug to filter which solution type is but I wonder if we should add a field "type" in Strapi instead. Maybe using the slug for this is not totally safe. Thoughts?

travjenkins: No - I think this is fine. We just need to make sure folks know they cannot change these on a whim (which they don't do anyway).

src/components/HeaderNavbar/Solutions/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+import {
+    useCaseSolutions,
+    industrySolutions,
+    technologySolutions,
+} from './items';
+
+const HeaderNavbarSolutions = ({ active, setActive }) => {
+    const isMobile = useMediaQuery('(max-width:1142px)');
+
+    const onClick = (ev: { preventDefault: () => void }) => {
```
travjenkins, line 21: Should be safe to use React`s typing here `MouseEvent`

src/components/HeaderNavbar/Solutions/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+                onMouseEnter={onMouseEnter}
+                sx={accordionSummaryStyles}
+            >
+                <MenuLink title="Solutions" active={active} />
+            </AccordionSummary>
+            <AccordionDetails className={accordionDetails}>
+                <Card>
+                    <CardItem title="USE CASES" items={useCaseSolutions} />
+                    <CardItem title="INDUSTRY" items={industrySolutions} />
+                    <CardItem title="TECHNOLOGY" items={technologySolutions} />
```
travjenkins, line 50: I feel like we should start using styling more for upper casing content. Thoughts?

src/components/HeaderNavbar/Solutions/items.tsx
```diff
@@ -0,0 +1,92 @@
+import DocIcon from '../../../svgs/document.svg';
+import PieChartIcon from '../../../svgs/pie-chart.svg';
+import NoCodeIcon from '../../../svgs/no-code.svg';
+
+const iconSize = 15;
```
travjenkins, line 5: let's add an `iconColor` const as well for storing `#5072EB`

src/components/HeaderNavbar/Solutions/items.tsx
```diff
@@ -0,0 +1,92 @@
...
+const AdaptedPieChartIcon = () => (
+    <PieChartIcon width={iconSize} height={iconSize} color="#5072EB" />
+);
+const AdaptedNoCodeIcon = () => (
+    <NoCodeIcon width={iconSize} height={iconSize} color="#5072EB" />
+);
+
+const urlPrefix = '/solutions';
+
+export const useCaseSolutions = [
```
travjenkins, line 19: Just calling out (no change is _truly_ needed) `useCaseSolutions` is a const name that you have used. This could make looking up where this public const is used a little complex.

image

src/components/HeaderNavbar/CardItem/index.tsx
```diff
@@ -14,9 +14,18 @@ import {
...
 
-const ItemLink = ({ name, description, Image, to }) => {
-    const LinkElement: any = to[0] === '/' ? Link : OutboundLink;
-    const linkProps = to[0] === '/' ? { to } : { href: to, target: '_blank' };
+interface Item {
+    name: string;
+    description?: string;
+    to: string;
+    Image?: () => JSX.Element;
+}
```
travjenkins, line 22: This typing should be shared and then the arrays of pages in `src/components/HeaderNavbar/Solutions/items.tsx` can be typed

gatsby-node.ts
Brenosalv: Okay, I'll let them know.

src/components/Solutions/Sections/Highlights/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+            <Container className={wrapper} isDarkTheme={isDarkTheme} isVertical>
+                <Container className={noPadding}>
+                    <div className={sectionText}>
+                        <h2>
+                            <span>{data.sectionTitle.highlightedText}</span>{' '}
+                            {data.sectionTitle.normalText}
+                        </h2>
+                        <p>{data.description}</p>
+                    </div>
+                    {isDarkTheme ? <ImageDark /> : <ImageLight />}
```
travjenkins, line 34: thoughts for future feature - adding light/dark mode support into the `svg`s with a common class we use.

src/images/technology-solutions-template/hero-image.png
travjenkins, line 1: If these two images are not used in link sharing... we probably wan to just make the background the image and then pull the two different inner circles out

travjenkins (commented):
overall looks good

some small changes called out

going to test the preview now

src/components/HeaderNavbar/Solutions/index.tsx
Brenosalv: Yes, I was thinking about not doing this to avoid more code, but it makes sense to add one more safety layer.

src/components/HeaderNavbar/Solutions/items.tsx
Brenosalv: Thanks for spotting this. I changed the menu consts from `useCaseSolutions` to `useCaseSolutionItems` format.

src/components/HeaderNavbar/CardItem/index.tsx
Brenosalv: Thanks for spotting this.

src/components/Solutions/Sections/Highlights/index.tsx
Brenosalv: I tried to do this but it's not simple, so okay I will leave this for the future.

travjenkins (approved):
lgtm

Issue #628, opened 2025-01-09
New connector Dekaf is causing the build to fail.


Brenosalv:
**Bug**
New connector Dekaf is causing the build to fail.

**Expected**
Filter the connectors from the `normalizeConnectors` function.

**Screenshots**
<img width="1019" alt="image" src="https://github.com/user-attachments/assets/1f320bb6-b890-4578-a3c6-6d7cb3d6b77b" />

PR #629, opened 2025-01-09
Filter connectors removing the Dekaf connector

Brenosalv:
#628

## Changes

-   Filter the connectors removing the Dekaf connector.

It was not possible to filter connectors removing Dekaf through GraphQL queries, so it was done through the `normalizeConnector` function reaching all the function instances.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -14,7 +14,7 @@ const CONNECTOR_IMAGE_RE = /(source|materialize)-([a-z0-9\-]+)/;
 export const normalizeConnector = (connector: any) => {
     if (connector.imageName === 'ghcr.io/estuary/dekaf-generic') {
         // Exclude Dekaf connector through the imageName because the id might change periodically.
-        return [];
+        return undefined;
     }
     if (!connector) {
         return connector;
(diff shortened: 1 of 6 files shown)


src/utils.ts
```diff
@@ -12,6 +12,10 @@ import {
 const CONNECTOR_IMAGE_RE = /(source|materialize)-([a-z0-9\-]+)/;
 
 export const normalizeConnector = (connector: any) => {
+    if (connector.id === '0f:df:77:d9:d8:77:bc:00') {
+        // Exclude Dekaf connector
+        return [];
+    }
```
travjenkins, line 18: We don't want to check the `id` this closely as that _could_ change.... can we just see if the image is `ghcr.io/estuary/dekaf-generic`. 

This can also change but a bit easier to manage and add more to if needed.

Brenosalv, line 18: I tried to return `null` or `undefined` but it caused another error. Returning an array worked.

Brenosalv: Yes, the Dekaf properties are:
image

I can change to filter using the image.

Brenosalv: Nevermind, now it's working with `undefined`.

travjenkins (approved):
lgtm - considering we returned the `connector` if it wasn't there we probably should have always been filtering like this.

PR #630, opened 2025-01-10
Rename/remove several comparison matrix fields

aeluce:
## Changes

We’d like to make some updates to streamline our comparison matrix. This PR provides a bit of initial setup for that, removing a few fields and renaming others. It looks like more drastic changes, such as adding fields or moving them between sections, will require Strapi admin work to update the underlying API.

## Tests / Screenshots

-   Confirmed that the updated matrix layout appeared as expected. For example, compare the current use case fields:

old_matrix

With the updated version (no ODS Replication row; updated field labels):

new_matrix

### DIFF
--- a/shared.ts
+++ b/shared.ts
@@ -124,7 +124,6 @@ export interface Vendor {
     logo: any;
     slugKey: string;
     useCases: {
-        odsReplication: HasFeature;
         databaseReplication: string;
         dataMigration: HasFeature;
         dataIntegration: HasFeature;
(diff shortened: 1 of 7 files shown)


danthelion (approved):
Looks good, thank you!

travjenkins (approved):
At some time we'll probably want to open an issue and get Strapi updated if we are not going to use this data. However, probably best to know 100% we won't ever want it again.

PR #669, opened 2025-02-05
adding an initial setting for report only CSP

travjenkins:
## Changes

https://github.com/estuary/security/issues/217

- Adding a report only CSP header that is totally wide open but something so we can start seeing the impact

## Tests / Screenshots

- N/A


Brenosalv (approved):
Lgtm

PR #675, opened 2025-02-07
Fix: Remove flex display from paragraphs

aeluce:
## Changes

-   Paragraphs in case studies are treating components like links as separate columns when using Flex display. Removing this styling displays paragraphs more normally, with in-line links.

## Tests / Screenshots

Tested locally to confirm. Currently, paragraphs in case studies look like this when they contain links:
de_link_columns

With this fix, they will now appear as:
de_link_fixed

### DIFF
--- a/src/templates/success-story/styles.module.less
+++ b/src/templates/success-story/styles.module.less
@@ -185,7 +185,6 @@
 
         p {
             line-height: 30px;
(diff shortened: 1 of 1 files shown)


travjenkins (approved):
lgtm

Issue #676, opened 2025-02-10
Update success story list


danthelion:
on page: https://estuary.dev/success-stories/

in nav panel: replace true w david energy. if possible generate list automatically sorted by date desc

Brenosalv:
@danthelion This is how descending order would look like in menu:

<img width="683" alt="Image" src="https://github.com/user-attachments/assets/63d14c0b-df5c-488a-a067-0acfb9e8ac64" />

Will we add descending order to the grid from https://estuary.dev/success-stories/ as well or only the menu?

PR #678, opened 2025-02-11
Update success story menu list and add redirects for True Platform

Brenosalv:
#676 

## Changes

-   Hardcode order of success stories in the resources menu;
-   Add Firebase redirects to the True Platform URLs

## Tests / Screenshots

image

image


travjenkins (approved):
lgtm

PR #683, opened 2025-02-12
Replace hardcoded text from key features section

Brenosalv:
## Changes

-   Replace hardcoded text from key features section with content from Strapi;
-   Remove unused Tag component.

## Tests / Screenshots

image


src/components/Solutions/Sections/KeyFeatures/index.tsx
```diff
@@ -34,19 +34,19 @@ const KeyFeatures = ({ data, isDarkTheme = false }: KeyFeaturesProps) => {
                     <p>{data.description}</p>
                 </div>
                 <Card
-                    description="Real-time transformations at the source or destination."
+                    description={data.keyFeatureItems.strapi_json_value[0]}
                     svg={<KeyFeature2 color={iconColor} />}
```
travjenkins, line 38: Should we validate that there are always three items?

Brenosalv: We could accept more items but I'm considering the layout was designed to accept only three items. Three boxes with different colors. Do you think we should accept more items?

travjenkins: I mean that we always get 3 items and not 2 or 1

Brenosalv: Got it. I adapted the code to accept from 0 to 3 items:

0 item:
image

1 item:
image

2 items:
image

3 items:
image

Brenosalv: We can make it scalable accepting more than 3 items, but I think the section would be bigger than expected.

travjenkins (approved):
lgtm

src/components/Solutions/Sections/KeyFeatures/index.tsx
travjenkins: That looks great! Kinda nice to not have to _always_ have it be 3 items.

PR #687, opened 2025-02-13
Upgrade `gatsby` to latest

travjenkins:
## Changes

- ran `yarn upgrade gatsby`
- Broke out `build` and `preview` in the workflow
- Skip `build` while in `draft` state

### `draft` with proper label still skips these two steps
image


.github/workflows/deploy-preview.yml
```diff
@@ -33,15 +32,19 @@ jobs:
                   key: ${{ runner.os }}-gatsby-build-preview-${{github.head_ref}}-${{ github.run_id }}
                   restore-keys: |
                       ${{ runner.os }}-gatsby-build-preview-${{github.head_ref}}-
-            - run: yarn install
+
+            - run: yarn install --frozen-lockfile
```
travjenkins, line 36: safer to run with a frozen lock file

.github/workflows/deploy-preview.yml
```diff
@@ -52,7 +55,9 @@ jobs:
                   GATSBY_DB_HOST: ${{ secrets.GATSBY_DB_HOST }}
                   GATSBY_DB_PORT: ${{ secrets.GATSBY_DB_PORT }}
                   GATSBY_DB_NAME: ${{ secrets.GATSBY_DB_NAME }}
+
             - uses: FirebaseExtended/action-hosting-deploy@v0
+              if: contains(github.event.pull_request.labels.*.name, 'generate-pr-preview')
```
travjenkins, line 60: We only want to block the deploy portion and not the build portion

.github/workflows/deploy-preview.yml
```diff
@@ -2,26 +2,32 @@ name: Deploy to Firebase Preview
...
     pull_request:
-    label:
-        types: [created, deleted, edited]
+        types:
+            [
+                opened,
+                synchronize,
+                reopened,
+                labeled,
+                unlabeled,
```
travjenkins, line 11: Get when labels change

.github/workflows/deploy-preview.yml
```diff
@@ -2,26 +2,32 @@ name: Deploy to Firebase Preview
...
+                ready_for_review,
```
travjenkins, line 12: Make sure we know when something is marked as ready for review

.github/workflows/deploy-preview.yml
```diff
@@ -2,26 +2,32 @@ name: Deploy to Firebase Preview
 
 on:
     pull_request:
-    label:
-        types: [created, deleted, edited]
+        types:
+            [
+                opened,
+                synchronize,
+                reopened,
```
travjenkins, line 9: Basic PR operations

.github/workflows/deploy-preview.yml
```diff
@@ -33,16 +39,22 @@ jobs:
...
+
             - name: Run Prettier
               run: yarn format
               env:
                   CI: true
-            - run: yarn run build --log-pages
+
+            - name: Run build
+              run: yarn run build --log-pages
+              if: ${{ github.event.pull_request.draft != true }}
```
travjenkins, line 57: We can skip builds until something is ready to review

.github/workflows/deploy-preview.yml
```diff
@@ -52,11 +64,14 @@ jobs:
                   GATSBY_DB_HOST: ${{ secrets.GATSBY_DB_HOST }}
                   GATSBY_DB_PORT: ${{ secrets.GATSBY_DB_PORT }}
                   GATSBY_DB_NAME: ${{ secrets.GATSBY_DB_NAME }}
+
             - uses: FirebaseExtended/action-hosting-deploy@v0
+              if: ${{ github.event.pull_request.draft != true && contains(github.event.pull_request.labels.*.name, 'generate-pr-preview') }}
```
travjenkins, line 69: Only preview if we are not in draft AND the label is there

package.json
```diff
@@ -21,7 +21,7 @@
     "dotenv": "^16.0.3",
     "embla-carousel": "^8.2.0",
     "embla-carousel-react": "^8.2.0",
-    "gatsby": "^5.13.3",
+    "gatsby": "^5.14.0",
```
travjenkins, line 24: Upgrading package as well

.lighthouserc.js
travjenkins, line 1: This is no longer needed

.github/workflows/deploy-preview.yml
Brenosalv: Is it really needed? Asking because we already have the built-in feature "request review" from GitHub.

.github/workflows/deploy-preview.yml
Brenosalv: I’m not sure about skipping builds for draft PRs since it might still be necessary to build and preview even before it’s ready for code review - for example, when the code is a work in progress, but we need to see what’s been done so far.

Brenosalv:
The build is failing for the same reason as yesterday:
image

package.json
Brenosalv, line 24: I tested new version locally and it seems to work well.

Brenosalv (dismissed):
Lgtm - I left a comment about the draft, but it's on your call.

.github/workflows/deploy-preview.yml
travjenkins: Honestly - that should be done on your local. And then if you need the build and preview (like you need someone to interact with the preview to test) you can just open up the PR.

This is also easier than adding another label of `generate-pr-build` which I considered.

Brenosalv: Okay, this way also works for me.

travjenkins: If this becomes an issue we can add the new label... I just wanna try this first. The main driving factor is all the builds are hammering Strapi and that is one of the reasons some of the builds are failing.

PR #693, opened 2025-02-20
Removing listening to changes on Connectors - `build perf` 

travjenkins:
## Changes

- Stop listening to connector table changes
- Added in some commented out code that would allow us to check for connectors at a certain cadence. Right now those updates are hammering our Strapi endpoint. Also, the changes happening are not too often related to what we need to worry about. For instance - if a schema changes we do not actually care.

## Tests / Screenshots

- N/A

### DIFF
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -2,15 +2,6 @@ name: Deploy to Firebase Preview
 
 on:
     pull_request:
-        types:
-            [
-                opened,
-                synchronize,
-                reopened,
-                labeled,
-                unlabeled,
-                ready_for_review,
-            ]
 
 concurrency:
     group: preview-${{ github.ref }}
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy-preview.yml
```diff
@@ -66,7 +56,7 @@ jobs:
                   GATSBY_DB_NAME: ${{ secrets.GATSBY_DB_NAME }}
 
             - uses: FirebaseExtended/action-hosting-deploy@v0
-              if: ${{ github.event.pull_request.draft != true && contains(github.event.pull_request.labels.*.name, 'generate-pr-preview') }}
+              if: ${{ contains(github.event.pull_request.labels.*.name, 'generate-pr-preview') }}
```
travjenkins, line 59: We will now expect the label to be on the PR when it is opened.

Brenosalv (approved):
Lgtm

PR #695, opened 2025-02-20
Make blog post search better

travjenkins:

## Changes

- Altering how we weight different things in search
- Always search lower case to keep things more consistent
- Inform users when not results were found

### Now we will inform user when we cannot find blog posts
image

### `gcs` search returns less results but much more accurate to query
image


gatsby-config.ts
```diff
@@ -532,7 +533,7 @@ const cfg: GatsbyConfig = {
                 // List of keys to index. The values of the keys are taken from the
                 // normalizer function below.
                 // Default: all fields
-                index: ['title', 'searchable_tags'],
+                index: ['slug', 'title', 'searchable_tags'],
```
travjenkins, line 536: the `slug` sometimes contains some interesting values that might match what the user is searching for

danthelion (approved):

PR #699, opened 2025-02-21
Filter `connectors` better and make connector search page safer

travjenkins:
## Changes

- Add handling to remove connectors like `kelkoo`
- Make connector logo loading more safe since there is a chance they'll be missed

#### Prod
image
#### Local
image


src/components/Connectors/index.tsx
```diff
@@ -65,48 +65,66 @@ const ConnectorCard = ({
...
-                    <div>
-                        <p className={connectorPostCardRecommended}>
-                            RECOMMENDED
-                        </p>
-                    </div>
-                ) : null}
-                {showType ? (
-                    <>
-                        {recommended ? <div style={{ flexBasis: 4 }} /> : null}
+}: ReturnType<typeof normalizeConnector> & { showType?: boolean }) => {
```
travjenkins, line 68: This didn't change much but looks like it did because I added a explicit `return`.

src/components/Connectors/index.tsx
```diff
@@ -240,14 +252,16 @@ export const Connectors = ({
...
                     {(query.length > 0 ? results : mappedConnectors).map(
-                        (connector) => (
-                            <ConnectorCard
-                                key={connector.id}
-                                {...connector}
-                                logo={logosByConnectorId[connector.id]}
-                                showType={showAllConnectors}
-                            />
-                        )
+                        (connector) => {
```
travjenkins, line 255: another explicit `return`

src/components/ConnectorsLink/index.tsx
travjenkins, line 1: We now check this in `normalizeConnector`.

src/templates/connection.tsx
travjenkins, line 1: This should never happen - but being safe

src/utils.ts
travjenkins, line 1: We need to add all the filter into the normalizer so that things like search results do not get fed wrong connectors

Brenosalv (dismissed):
Lgtm

Brenosalv (approved):
Lgtm

Issue #623, opened 2025-01-08
Move blogs to /blog


danthelion:
Moving to a /blog path can definitely have benefits, like improving URL structure and helping search engines and users better understand our content hierarchy. There might be some ranking or performance fluctuations but they’re typically short-lived if handled properly. I believe it’s worth doing for the long term...
Here’s what we need to keep in mind:

1. Set up permanent (301) redirects carefully from the old URLs to the new /blog paths.
2. Update all internal links to point directly to the /blog URLs in content, menus, and footers. (Can this be done in bulk?)
3. Update  XML sitemap with the new /blog URLs.
4. Verify canonical tags on the new blog pages point to the correct /blog URLs.
5. Verify that schema markups is still intact and reflects the new paths.
6. Map old redirect URLs to their new /blog URLs to avoid redirect chains (e.g., A → B → C).

danthelion:
discussion: https://estuaryworkspace.slack.com/archives/C079QLEMWKA/p1736274606031319

travjenkins:
I think we might want to wait for the Next JS migration to do this.

Brenosalv:
Yes, with Gatsby we would need to use `onPostBuild` API on `gatsby-node.js` to fetch the slugs and add the redirects with them. In Next.js it will be different. So yeah, it might make sense to wait in case it's not so urgent.

PR #704, opened 2025-02-24
Moving blog posts under '/blog' prefix

travjenkins:

## Changes

https://github.com/estuary/marketing-site/issues/623
- [x] generate prefixed slugs
- [x] update `firebase.json`
- [x] add a check for new blogs so we can keep an eye for a little while

### What a build failure would look like (faked by just checking if the name had length greater than 1)
image

### DIFF
--- a/gatsby-node.ts
+++ b/gatsby-node.ts
@@ -355,6 +355,7 @@ const createBlogs: CreateHelper = async (
                         );
                     }
 
+                    // TODO - remove this after about one month
                     console.log(
                         `blogPost:redirect:${createdAfterSwitch ? 'new' : 'old'}`,
                         newPath
--- a/gatsby-node.ts
+++ b/gatsby-node.ts
@@ -334,8 +334,32 @@ const createBlogs: CreateHelper = async (
                     const nextPostId =
                         index === posts.length - 1 ? null : posts[index + 1].id;
 
+                    const oldPath = post.Slug;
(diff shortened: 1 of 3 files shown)


gatsby-node.ts
```diff
@@ -166,20 +166,16 @@ const createVendorCompare: CreateHelper = async (
...
-    { actions: { createPage, createRedirect }, graphql, reporter }
+    { actions: { createPage }, graphql, reporter }
 ) => {
     const startTime = performance.now();
     console.log(`Creation:Start:${name}`);
 
-    createRedirect({
-        fromPath: '/blogs',
-        toPath: '/blog',
-    });
```
travjenkins, line 177: We do not seem to consume this in anyway and just manage it manually in `firebase.json`.

gatsby-node.ts
```diff
@@ -334,8 +331,38 @@ const createBlogs: CreateHelper = async (
...
+                        throw new Error(
+                            `Blog post has slug that would overlap with search tabs: ${post.id}`
+                        );
+                    }
+
+                    // TODO - remove this after about one month
+                    console.log(
+                        `blogPost:redirect:${createdAfterSwitch ? 'new' : 'old'}`,
+                        newPath
+                    );
```
travjenkins, line 362: I think eventually we don't need to log this but for awhile we should keep an eye on the build

Brenosalv (approved):
Lgtm

Issue #707, opened 2025-02-24
Build failing due to connector's null logo


Brenosalv:
**Bug**
Build is failing probably because Gatsby is trying to access the `gatsbyImageData` from a connector's null logo. 

**Expected**
Prevent Gatsby to do this and then solve the build failing.

**Screenshots**

<img width="881" alt="Image" src="https://github.com/user-attachments/assets/54423258-3629-4a61-9945-988a00ea3fa1" />

PR #708, opened 2025-02-25
Solve build failing due to connectors null logo

Brenosalv:
#707 

## Changes

-   Prevent Gatsby from accessing `gatsbyImageData` of null logo property.
-   Add placeholder icon when logo is null on integration and connector pages and the connector filter icons.

## Tests / Screenshots


Brenosalv:
Not sure about the icon:
image

Brenosalv:
@travjenkins Recently the Monday logo was added with this size and format:
image
image

Is it possible to replace it with the same size of the other connector logos? Just removing the "monday" and keeping the image of the logo would work.

src/components/ConnectorsLink/index.tsx
travjenkins, line 1: This should be removed. We are already filtering this out up above

image

src/components/Integration/Hero/index.tsx
```diff
@@ -37,12 +37,12 @@ export interface HeroProps {
...
-    );
-    const destinationConnectorLogo = getImage(
-        destConnector.logo?.childImageSharp?.gatsbyImageData
-    );
+    const sourceConnectorLogo = sourceConnector.logo
+        ? getImage(sourceConnector.logo?.childImageSharp?.gatsbyImageData)
+        : null;
+    const destinationConnectorLogo = destConnector.logo
+        ? getImage(destConnector.logo?.childImageSharp?.gatsbyImageData)
+        : null;
```
travjenkins, line 45: If `destConnector.logo?.childImageSharp?.gatsbyImageData` is optionally chained then the ternary should probably just check the full chain to get safe.

travjenkins (commented):
One change needs to happen.

src/components/ConnectorsLink/index.tsx
Brenosalv: Yes, this filter was removed on the first commit.

Brenosalv:
I asked this icon to Polina and will replace the current ones once I have it:
image

travjenkins (approved):
lgtm

Issue #713, opened 2025-02-26
Add og:image for social previews to case studies


danthelion:
<img width="843" alt="Image" src="https://github.com/user-attachments/assets/0bbec161-60c5-4d8e-bbe6-0fd6b9ca6996" />

PR #718, opened 2025-02-28
Fix social preview image from success story page

Brenosalv:
#713 

 Open graph image from success story pages not being displayed.
 
## Changes
-   Use the right property from the fetched meta image.


travjenkins (approved):
Lgtm

Issue #716, opened 2025-02-27
Numbering Bug Beyond 9


SourabhEstuary:
There's a numbering issue in our blogs. After 9, the numbers don’t display correctly - they show up with an extra period (like ".0", ".1", ".2"). Can you look into why this is happening and fix it so that numbers display correctly after 9?

Sample blogs : 
- https://estuary.dev/blog/elt-tools/
- https://estuary.dev/blog/real-time-data-integration-postgres-s3-data-lake/

![Image](https://github.com/user-attachments/assets/131cc8b0-5049-4a3b-b5cc-d77870dc1e2a)

![Image](https://github.com/user-attachments/assets/44049ca3-0524-4213-bf19-a1d54e3c588f)

SourabhEstuary:
Got it! It seems like a styling issue. Could you check if this can be fixed to ensure full visibility?

Brenosalv:
Yes, it was a styling issue. I will fix this.

PR #736, opened 2025-03-07
Fix list element marks cut off in blog post body

Brenosalv:
#716

## Changes

-   Fix list element markers cut off on blog post body;

I fixed this bug but leveraged this PR to solve other bugs as well:
-   Fix blog post table not scrolling as expected;
-   Fix comparison table header with extra spacing on the top;
-   Fix blog post sidebar's wrong width on the comparison page;
-   Fix mobile and desktop menus conflict - both were opening on first page load on mobile.

## Tests / Screenshots

Now the list item markers are being totally displayed:
image

Another bug - off topic:
image
Fixed:
image

### DIFF
--- a/src/components/ArticleSidebar/styles.module.less
+++ b/src/components/ArticleSidebar/styles.module.less
@@ -1,12 +1,12 @@
 .container {
   display: flex;
   flex-direction: column;
-  max-width: 256px;
+  min-width: 256px;
   width: 100%;
   gap: 24px;
   height: fit-content;
 
-  @media (max-width: 1150px) {
+  @media (max-width: 1024px) {
     max-width: 100%;
   }
 
@@ -25,7 +25,7 @@
 .shareArticleDesktop {
   display: flex;
 
-  @media (max-width: 1150px) {
+  @media (max-width: 1024px) {
     display: none;
   }
 }
@@ -33,7 +33,7 @@
 .tableOfContentsWrapper {
   overflow: auto;
 
-  @media (max-width: 1150px) {
+  @media (max-width: 1024px) {
     margin-bottom: 32px;
   }
 }
--- a/src/components/EtlToolsXvsYPage/Comparison/styles.module.less
+++ b/src/components/EtlToolsXvsYPage/Comparison/styles.module.less
@@ -51,7 +51,7 @@
 
     thead th:not(:first-of-type) {
       position: sticky;
-      top: 116px;
+      top: 80px;
       padding: 0;
     }
 
--- a/src/components/Header/index.tsx
+++ b/src/components/Header/index.tsx
@@ -1,7 +1,7 @@
 import clsx from 'clsx';
 import { Link } from 'gatsby';
 import { useEffect, useMemo, useRef, useState } from 'react';
-import { Slide, useMediaQuery } from '@mui/material';
+import { Slide } from '@mui/material';
 import MenuIcon from '@mui/icons-material/Menu';
 import ColoredLogo from '../../svgs/colored-logo.svg';
 import GithubIcon from '../../svgs/github-outline.svg';
@@ -34,6 +34,8 @@ import {
     globalHeaderMobileMenuWrapper,
     globalHeaderMobileMenuButton,
     headerSocialIcon,
+    menuContentDesktop,
+    menuContentMobile,
 } from './styles.module.less';
 
 const slideTimeout = 150;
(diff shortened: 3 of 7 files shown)


travjenkins (approved):
lgtm

plugins/estuary-rehype-transformers/src/index.js
```diff
@@ -9,18 +9,58 @@ import { unified } from 'unified';
...
 
+// Checks if a node is a GatsbyImage wrapper.
+function isGatsbyImageWrapper(node) {
+    if (node.tagName !== 'div' || !node.properties) return false;
+    const hasDataAttr = node.properties['data-gatsby-image-wrapper'] !== undefined;
+    const hasClass =
+        node.properties.className &&
+        (Array.isArray(node.properties.className)
+            ? node.properties.className.includes('gatsby-image-wrapper')
+            : node.properties.className.includes('gatsby-image-wrapper'));
```
travjenkins, line 22: To always ensure `gatsby-image-wrapper` is the same class in this function we should move that out into a `const`

travjenkins, line 22: What is driving checking if it is an array here? Since string and array both use the `includes` function to check for contents?

src/components/BlogPost/index.tsx
```diff
@@ -220,40 +221,41 @@ const BlogPost = ({
...
-                                );
-
                             const authorBio = author?.bio.data.bio;
 
                             const authorSocialLinks = author?.socials;
 
                             return (
-                                <>
-                                    <div key={index} className={authorInfo}>
+                                <Fragment key={author.slug || index}>
```
travjenkins, line 229: This is technically okay... but I do not think we should ever use this kind of `key` if we can help it.

If we just use the slug and index in a single string it should stay unique.

src/templates/blog/index.tsx
```diff
@@ -184,7 +184,7 @@ const BlogIndex = ({
                     ) : null}
 
                     {postsToRender.map((post) => (
-                        <BlogPostCard key={post.id} {...post} />
+                        <BlogPostCard key={post.slug} {...post} />
```
travjenkins, line 187: what drove this change?

travjenkins (dismissed):
Overall - looks alright to me. My call outs are tiny and not needed to be worried about before merging.

plugins/estuary-rehype-transformers/src/index.js
Brenosalv: It's just for future proof. It's possible the className comes in as only a class or as an array of classes, so it's good to check this to avoid possible future type issues. This check guarantees that we correctly detect `gatsby-image-wrapper` regardless of how the class names are stored.

Even though the code ends up calling .includes in both cases, the check ensures that if className is an array, we’re using the array method, and if it’s a string, we’re using the string method.

src/components/BlogPost/index.tsx
Brenosalv: Yes, we can use the `id` here and do what you said:

`${author.id || 'author-id'}-${index}`

src/templates/blog/index.tsx
Brenosalv: The authors' `id` were coming `undefined` and then not unique, causing the warning. So I switched to `slug`, but I just noticed that happened because the `id` was not included in the GraphQL query. As it's fixed, I will switch back to using the `id`: `${author.id}-${index}` .

plugins/estuary-rehype-transformers/src/index.js
```diff
@@ -15,12 +15,15 @@ const LANG_RE = /hljs language\-(.*)/;
 function isGatsbyImageWrapper(node) {
     if (node.tagName !== 'div' || !node.properties) return false;
     const hasDataAttr = node.properties['data-gatsby-image-wrapper'] !== undefined;
-    const hasClass =
+
+    const GatsbyImageWraperClassName = 'gatsby-image-wrapper';
```
travjenkins, line 19: Sorry to be nit picky... but this needs to start with a lowercase letter.

Brenosalv: Sorry for lacking attention on this 🤦🏻‍♂️

travjenkins: No worries man.

travjenkins (dismissed):
lgtm

travjenkins (approved):
lgtm

Issue #520, opened 2024-10-08
Replace Current Pop-up Banner for "Tutorial" Category Blogs


SourabhEstuary:
**Objective:** 

To replace the current exit-intent pop-up banner with a new banner that is specifically targeted for visitors of "Tutorial" blog pages. The new banner will showcase Estuary's USP, encouraging users to try Estuary for moving data from "X" to "Y."

**Purpose:**

The main purpose of this pop-up is to make a final attempt to attract users to register on the website and try Estuary for moving data from "X" to "Y." The pop-up will activate when a user’s cursor indicates intent to close the browser tab or window. It will present a clear and compelling message about Estuary’s capabilities, driving them to take action and register for the platform.

**Message to Showcase on Pop-Up**

**Heading:** Instantly Move Your Data – No Code Needed
**Subheading:** Move Data from Any Source to Any Destination in Minutes – No Coding Required

**Key points:**

- Real-time and batch integration in one unified platform
- Enterprise-ready support for all private networking environments
- Load data into multiple destinations simultaneously
- Proven to scale beyond 7GB+/sec, supporting high-volume data

**Button 1:** Try Free for 30 Days  (Primary Button - Should be in highlighted and bold colour)
**Button 2:** Contact Us

**Text below button:** No credit card required

Pop-up example:
![image](https://github.com/user-attachments/assets/47493a3f-8183-4afa-bc05-c34cc6cfa522)

Brenosalv:
@altitude-lucas Can we reuse UI elements/components we already have on other pages or will we need a customized design for this? Asking because we have the design label in this issue.

altitude-lucas:
@Brenosalv I think it's okay to re-use what we have existing, if there are any specific design elements we're missing we can ask Polina for them but in this case we have most of what we need. @danthelion FYI

danthelion:
This is good to go from my side, let's put it in the backlog @SourabhEstuary

Brenosalv:
@danthelion We need an image to put on the right of the banner. Could you choose one with Polina?

SourabhEstuary:
Hey @danthelion .. need your help on image finalization for the pop-up...

danthelion:
moving to todo

Brenosalv:
@danthelion @SourabhEstuary Will we keep the old popup for the other blog posts that are not Tutorial?

Brenosalv:
Here is a first version of the popup so you can provide a feedback - don't worry about the image, I will adjust it yet:

[...]

PR #740, opened 2025-03-12
Replace current blog post pop-up banner

Brenosalv:
#520

## Changes

-   Replace blog post pop-up.

## Tests / Screenshots

Desktop:
image

Mobile (<425px) - I decided to remove the image once on mobile because it gets too small and the text becomes hard to read:
image


src/components/BlogPostPopupModal/index.tsx
```diff
@@ -1,32 +1,40 @@
...
-    image: (
-        <StaticImage
-            src="../../images/blog-post-popup-image.png"
-            alt="Change data capture webinar"
-            placeholder="blurred"
-        />
-    ),
+    image: <FlowDiagram diagramCardsClassName={diagramCards} />,
+    footerText: 'No credit card required',
     version: '',
```
travjenkins, line 38: You need to add a version here. Since this is a new key we can start this right by always having a version.

src/components/BlogPostPopupModal/index.tsx
```diff
@@ -35,21 +43,27 @@ const SETTINGS = {
...
     //  alt="Fake alt text"
     //  placeholder="none"
     //  loading="eager"
     // >,
+    // footerText: 'Fake text',
     // version: `/blogPostPopup_YYYY/MM/DD`,
 };
 
-const STORAGE_KEY = `@estuary/closeBlogPostPopup${SETTINGS.version}`;
+const STORAGE_KEY = `@estuary/closeBlogPostPopupTutorial${SETTINGS.version}`;
```
travjenkins, line 66: If we update the `version` we're fine - but why is the name of this changing?

src/components/BlogPostPopupModal/index.tsx
```diff
@@ -103,28 +120,51 @@ const BlogPostPopupModal = () => {
...
-                <IconButton onClick={handlePopupClose}>
-                    <CloseIcon fontSize="large" />
-                </IconButton>
-            </div>
             <div className={leftColumn}>
+                <h2>{SETTINGS.title}</h2>
+                <p>{SETTINGS.description}</p>
+                <ul>
+                    {SETTINGS.keyPoints.map((keyPoint, index) => (
+                        <li key={index}>
```
travjenkins, line 128: The `key` cannot just be `index` - you need to include some kind of description in there as well.

src/components/BlogPost/index.tsx
```diff
@@ -89,7 +89,7 @@ const BlogPost = ({
             itemScope
             itemType="http://schema.org/Article"
         >
-            <SwoopingLinesBackground>
+            <SwoopingLinesBackground loading="eager">
```
Brenosalv, line 92: The background image was flashing the alternative text before loading and it was probably impacting the SEO. Fixed.

src/components/BlogPostPopupModal/index.tsx
Brenosalv: I wonder why we didn't have a version for the old pop-up. Now we still have only one pop-up since we removed the old one.

src/components/BlogPostPopupModal/index.tsx
Brenosalv: Previously I thought we would add a new pop-up for the tutorials only instead of just replacing the old one. So just forgot to undo this change. Thanks.

src/components/BlogPostPopupModal/index.tsx
Brenosalv: Thanks for spotting this.

src/components/BlogPost/index.tsx
travjenkins: Oh cool - I have wondered what was causing that area to get called out during scans. Great find!

src/components/BlogPostPopupModal/index.tsx
Brenosalv: I think `pop-up-key-point_${index}` is good.

travjenkins (approved):
lgtm

Issue #739, opened 2025-03-11
Reverse sort order on homepage success stories + add missing


danthelion:
missing: headset

<img width="1147" alt="Image" src="https://github.com/user-attachments/assets/7a1c99c9-0f8c-4880-9de3-0b99f2ab6aab" />

danthelion:
hardcode first:

forward, headset, prodege

rest: descending based on publication date

Issue #732, opened 2025-03-06
Sort success stories by newest first


danthelion:
on success stories page

danthelion:
first 3: forward, headset, prodege
rest: sort desc by date

PR #742, opened 2025-03-12
Reverse sort order success stories on homepage and success stories page

Brenosalv:
#739 #732 

## Changes

-   Reverse sort order success stories on homepage and success stories page;
-   Remove limit of the homepage success stories so that the Headset success stories shows up;
-   Sort by createdAt descendent;
-   Add forward, headset and prodege as first success stories. 

## Tests / Screenshots

Homepage:
image

All success stories page:
image

### DIFF
--- a/shared.ts
+++ b/shared.ts
@@ -206,3 +206,23 @@ const getVendorLinkId = (vendorName?: string) => getSlugifiedText(vendorName);
 
 export const getVendorsLinkId = (vendorName1?: string, vendorName2?: string) =>
     `${getVendorLinkId(vendorName1)}-vs-${getVendorLinkId(vendorName2)}`;
+
+export const getSortedSuccessStories = (successStories) => {
+    const priorityOrder = ['forward', 'headset', 'prodege'];
+
+    const sortedSuccessStories = [...successStories].sort((a, b) => {
+        const indexA = priorityOrder.findIndex((keyword) =>
+            a.slug.includes(keyword)
+        );
+        const indexB = priorityOrder.findIndex((keyword) =>
+            b.slug.includes(keyword)
+        );
+
+        if (indexA === -1 && indexB === -1) return 0;
+        if (indexA === -1) return 1;
+        if (indexB === -1) return -1;
+        return indexA - indexB;
+    });
+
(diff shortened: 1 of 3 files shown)


shared.ts
```diff
@@ -206,3 +206,23 @@ const getVendorLinkId = (vendorName?: string) => getSlugifiedText(vendorName);
 
 export const getVendorsLinkId = (vendorName1?: string, vendorName2?: string) =>
     `${getVendorLinkId(vendorName1)}-vs-${getVendorLinkId(vendorName2)}`;
+
+export const getSortedSuccessStories = (successStories) => {
+    const priorityOrder = ['forward', 'headset', 'prodege'];
+
+    const sortedSuccessStories = [...successStories].sort((a, b) => {
```
travjenkins, line 213: Why are we spreading `successStories` here? Shouldn't this always be an array and ready to sory?

shared.ts
```diff
@@ -206,3 +206,23 @@ const getVendorLinkId = (vendorName?: string) => getSlugifiedText(vendorName);
 
 export const getVendorsLinkId = (vendorName1?: string, vendorName2?: string) =>
     `${getVendorLinkId(vendorName1)}-vs-${getVendorLinkId(vendorName2)}`;
+
+export const getSortedSuccessStories = (successStories) => {
+    const priorityOrder = ['forward', 'headset', 'prodege'];
```
travjenkins, line 211: This should be declared outside of this function so they aren't always re-declared

shared.ts
travjenkins, line 213: Same with this `sort` function being passed in. We should just declare that so the function is already ready.

travjenkins (dismissed):
Overall this is good. Just some small tweaks to help clean the code up a bit.

shared.ts
Brenosalv: I first did this to avoid changing the original array, but yeah… we ended up using it only on this line anyway, so no need to make a copy. Thanks.

shared.ts
Brenosalv: That's true. Fixed.

travjenkins (approved):
lgtm

PR #743, opened 2025-03-13
Centralize horizontally the blog post images

Brenosalv:
#733 

## Changes

-   Centralize horizontally the blog post images.

## Tests / Screenshots

Examples:
image

image

### DIFF
--- a/src/style.less
+++ b/src/style.less
@@ -284,6 +284,8 @@ span {
     .gatsby-image-wrapper,
     .raw-html-embed {
         display: flex !important;
+        flex-direction: column;
+        align-items: center;
         justify-content: center;
         width: fit-content;
         margin-left: auto;
--- a/src/style.less
+++ b/src/style.less
@@ -280,12 +280,19 @@ span {
         }
     }
 
-    figure, .gatsby-image-wrapper, .raw-html-embed {
+    figure,
+    .gatsby-image-wrapper,
+    .raw-html-embed {
         display: flex !important;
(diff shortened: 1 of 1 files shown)


src/style.less
```diff
@@ -279,6 +279,14 @@ span {
             margin-bottom: 0;
         }
     }
+
+    figure, .gatsby-image-wrapper {
+        display: flex !important;
```
Brenosalv, line 284: We had to force this because some of them come as `inline-block`.

Brenosalv:
@travjenkins What do you think about adding a default vertical margin to all the images? Maybe 32px. Images like the ones below have very small vertical margins.

image

src/style.less
Brenosalv: @travjenkins Could you share this blog post?

travjenkins: https://estuary.dev/blog/top-marketing-analytics-tools/

Brenosalv: Thanks

src/style.less
```diff
@@ -279,6 +279,23 @@ span {
             margin-bottom: 0;
         }
     }
+
+    figure,
```
travjenkins, line 283: I doubt we use `figure` anywhere else... but to keep the scope as small possible I think this should be `figure.image` just in case.

Brenosalv: I agree, it's safer

travjenkins (approved):
lgtm

Issue #737, opened 2025-03-07
Integrations page revamp


danthelion:
design doc: https://docs.google.com/document/d/1ryoUsmYwB-Vv3IUdhwhd6AM5uxis9IARhmYlrvtam6Y/edit?tab=t.0#heading=h.zdfzfedxt5xo

PR #748, opened 2025-03-25
Integrations page revamp

Brenosalv:
#737 

## Changes

-   Update sections and add others to the integration template page.


Brenosalv:
I added a solution to handle the connector logo images with `svg` format.

We were getting this issue for the CrateDB connector:

image

src/templates/connector/Hero/index.tsx
```diff
@@ -60,13 +54,21 @@ const Hero = ({ connector: { title, logo, type } }: HeroProps) => {
...
+                            />
+                        ) : logo?.childImageSharp?.gatsbyImageData ? (
                             <GatsbyImage
                                 alt={`${title} logo`}
-                                image={logoImage}
+                                image={logo.childImageSharp.gatsbyImageData}
+                                loading="eager"
                             />
                         ) : (
                             <ConnectorLogoPlaceholder connectorType={type} />
```
Brenosalv, line 74: We are accessing the `childImageSharp ` of the connector logo only when it's not an `svg`, but it's not solving the warning:

image

Brenosalv: We are not accessing the `childImageSharp` of the `svg` logo, but when fetching, the `svg` logo comes with `childImageSharp` and `publicURL`, not only `publicURL`:

image

I think we can't filter `svg` logo fetching `publicURL` only directly in the query.

Brenosalv: I think it would be fine to just suppress this warning, because it's not filterable through the query and we are already treating this after the fetching.

src/components/EtlToolsPage/Hero/index.tsx
```diff
@@ -28,17 +28,31 @@ const Hero = ({ vendors }: SectionOneProps) => {
...
+                                <GatsbyImage
+                                    key={vendor.id}
+                                    alt={`${vendor.name} logo`}
+                                    image={
+                                        vendorLogoLocalFile.childImageSharp
+                                            .gatsbyImageData
+                                    }
+                                    loading="eager"
+                                />
+                            ) : null
```
Brenosalv, line 53: I checked the image extension several times. We could create a component that returns an `img` or `GatsbyImage` depending on the extension being `svg` or not.

src/templates/connector/Hero/index.tsx
travjenkins: So I have updated this logo... we need to add something to strapi that prevents adding `svg` connector logos as we should just stick with `png` logos.

src/components/Connectors/index.tsx
```diff
@@ -78,7 +78,15 @@ const ConnectorCard = ({
...
                 <div className={connectorCardTop}>
-                    {logo?.childImageSharp?.gatsbyImageData ? (
+                    {logo?.extension === 'svg' ? (
+                        <img
+                            src={logo.publicURL}
+                            alt={`${title} Logo`}
+                            className={connectorLogoClassName}
+                            loading="eager"
+                            width={53}
+                        />
```
travjenkins, line 88: We can remove the handling of `svg` logos.

src/components/ConnectorsLink/index.tsx
```diff
@@ -72,8 +74,8 @@ const ConnectorsLink = ({
 
     const detailsHref = useMemo(() => {
         if (sourceId && destinationId) {
-            return `/integrations/${captureConnectors.find((c) => c.id === sourceId)?.slugified_name}-to-${
-                materializationConnectors.find((c) => c.id === destinationId)
+            return `/integrations/${captureConnectors.find((c) => c?.id === sourceId)?.slugified_name}-to-${
+                materializationConnectors.find((c) => c?.id === destinationId)
```
travjenkins, line 78: We should use `getIntegrationSlug` here

shared.ts
```diff
@@ -223,6 +252,31 @@ const sortSuccessStoriesByPriority = (a, b) => {
...
+    title: string;
+    logoUrl?: string;
+    recommended: boolean;
+    connectorTagsByConnectorIdList?: {
+        protocol: string;
+    }[];
+    slug: string | null;
+    type: ConnectorType;
+    slugified_name?: string;
+}
```
travjenkins, line 277: This looks like what is returned by the `normalizeConnector` none of these fields should be optional. They are 100% required for the site. We need to mark everything required and then if _anything_ is missing the connector is tossed out.

src/components/ConnectorsLink/index.tsx
```diff
@@ -85,15 +87,15 @@ const ConnectorsLink = ({
     const handleDestinationChange = (value) => setDestinationId(value);
 
     const sourceSelectItems = captureConnectors.map((c) => ({
-        id: c.id,
-        image: c.logo ? c.logo.childImageSharp.gatsbyImageData : null,
-        title: c.title,
+        id: c?.id,
+        image: c?.logo,
+        title: c?.title,
```
travjenkins, line 92: We should not have these be optional. Up above we filter out undefined connectors. So at that time we should just update the typing.

src/components/EtlToolsXvsYPage/Hero/index.tsx
```diff
@@ -20,7 +20,19 @@ interface SectionOneProps {
...
 const Hero = ({ vendors, xVendor, yVendor }: SectionOneProps) => {
+    const commonXvendorImageProps = {
+        alt: `${xVendor.name} logo`,
+        loading: 'eager' as LoadingType,
+    };
+
+    const commonYvendorImageProps = {
+        alt: `${yVendor.name} logo`,
+        loading: 'eager' as LoadingType,
+    };
```
travjenkins, line 34: We can remove `LoadingType` up above and then just type these as `ImgHTMLAttributes<HTMLImageElement>`

src/components/Grid/Card/index.tsx
```diff
@@ -129,31 +105,58 @@ const Card = ({
...
-                        {renderTags(data.tags)}
-                        {renderDateAndTime(data.updatedAt, readingTime)}
+                        {data.tags && data.tags.length > 0
+                            ? renderTags(data.tags)
+                            : null}
+                        {data.updatedAt && readingTime
+                            ? renderDateAndTime(data.updatedAt, readingTime)
+                            : null}
                     </div>
                 ) : null}
```
travjenkins, line 161: I think we're safe making this simpler... Both `renderTags` and `renderDateAndTime` return `null` if they are passing something false-y.

I think we can just always call those functions then have the logic be to check if either response is not null.

src/components/HeroSectionActions/RegisterCta.tsx
```diff
@@ -1,18 +1,22 @@
...
 
-function RegisterCta({ pageId }: RegisterCtaProps) {
+function RegisterCta({
+    pageId,
+    title = 'Start Building For Free',
+}: RegisterCtaProps) {
     return (
         <OutboundLinkFilled
-            id={`start-building-for-free-button/hero-section/${pageId}`}
+            id={`${getSlugifiedText(title)}-button/hero-section/${pageId}`}
```
travjenkins, line 15: We should just put the call to `getSlugifiedText` on the link components:

`/LinksAndButtons/LinkOutlined`
`/LinksAndButtons/LinkFilled`

That way we know they are always handled.

src/components/Integration/Faq/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+        };
+
+    return (
+        <section className={defaultWrapperGrey}>
+            <Container isVertical className={container}>
+                <h2>FAQ</h2>
+                <Wrapper>
+                    {faqs(sourceConnector)
+                        .filter((faq) => faq != null)
+                        .map(({ question, content }, index) => {
```
travjenkins, line 29: This requires two steps through to filter out. I think we're good just checking for `null` inside the `map` function.

src/templates/connector/Hero/index.tsx
Brenosalv: I created an issue for this.

src/components/EtlToolsPage/Hero/index.tsx
Brenosalv: Okay, I will undo this.

shared.ts
Brenosalv: I agree with you but at least the `slugified_name` has an undefined case:

image

src/components/ConnectorsLink/index.tsx
Brenosalv: You're right, thanks.

shared.ts
travjenkins: What I mean is that cannot be true. We cannot make a integrations page for connectors that have these cases.

So we should return undefined and throw an error when we find them so we can fix them.

src/components/Grid/Card/index.tsx
Brenosalv: Okay, I'll do that.

src/components/Integration/Faq/index.tsx
Brenosalv: Yeah, this way we avoid looping over the array twice.

src/components/Integration/Faq/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+                            const questionNumber = index + 1;
+                            return (
+                                <Accordion
+                                    key={questionNumber}
```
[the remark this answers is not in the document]
Brenosalv: I will pay more attention on this next time. Thanks.

shared.ts
Brenosalv: Got it! I will do this.

Brenosalv: Got it, we really always need this property as well.

Brenosalv: Actually the `shortDescription` and `longDescription` are optional because some of the connectors have short or long description, not both and it's not a problem. Currently, we prioritize using the long description on the integration FAQ page, and if that’s not available, we fall back to the short one. They are being used only on the FAQ on the integration page. If we don't have none of them, we just hide the question "What is [connector name]?".

Brenosalv: About `logoUrl`, I noticed locally that it's missing for PostgreSQL source and destination connectors.

Brenosalv: Also, I think throwing a warning and returning undefined when a property is missing instead of just throwing an error would fit better here.

Brenosalv: It was missing in a query from related integration links, now it's fixed. Throwing an error in case a property is missing really helps a lot, thanks!

src/components/Integration/Faq/index.tsx
```diff
@@ -24,26 +24,29 @@ const Faq = ({ sourceConnector }: FaqProps) => {
...
-                                    onChange={handleChange(
-                                        `question${questionNumber}`
-                                    )}
-                                >
-                                    {content}
-                                </Accordion>
-                            );
-                        })}
+                    {faqs(sourceConnector).map((faq, index) => {
+                        if (faq == null) {
```
travjenkins, line 28: I think we can just check `!faq` here

src/utils.ts
```diff
@@ -30,25 +30,57 @@ export const normalizeConnector = (
...
+    }
+
+    const type = connectorTagsByConnectorIdList[0]
+        ?.protocol as ConnectorType | null;
+
+    if (!type) {
+        throw new Error(
+            `Connector type is missing for connector with name: ${title['en-US']}`
+        );
+    }
```
travjenkins, line 67: Can we make these all the same and follow a pattern that maybe one day could be parsable. Also, just using the `id` is fine as that is how we're going to query the db anyway

`Error:connector:${id}:missing prop:Title`
`Error:connector:${id}:missing prop:imageName`
`Error:connector:${id}:missing prop:Slugified`
`Error:connector:${id}:missing prop:logoUrl`
`Error:connector:${id}:missing prop:Connector`

Brenosalv: I tried to create a separate `throwingError` function to reuse the error text but it's returning typescript issues because the null values are not being filtered when calling the `throwingError` function, so I will add the pattern but without this helper function.

travjenkins: sounds good - I'm fine with some duplication here since everything is in a single function.

PR #754, opened 2025-03-28
Testing experimental cache option

travjenkins:

## Changes

- No longer make source maps as we don't want them anyway
- Increase the node version to 20
- Removed `AVIF` support to reduce memory

## Tests / Screenshots

- Ran a lot of builds and looked at the preview when it worked
- Added and removed changes to see what seemed to help


.github/workflows/deploy-preview.yml
```diff
@@ -17,7 +17,11 @@ jobs:
 
             - uses: actions/setup-node@v1
               with:
-                  node-version: '18.x'
+                  node-version: '20.x'
```
travjenkins, line 20: The deployment to firebase will fail now as the CLI tool does not support `18` anymore.

gatsby-config.ts
```diff
@@ -356,6 +356,9 @@ const cfg: GatsbyConfig = {
     },
     // graphqlTypegen: true,
     plugins: [
+        {
+            resolve: 'gatsby-plugin-no-sourcemaps',
+        },
```
travjenkins, line 361: We do not need source maps and they are known to slow down the build

https://www.gatsbyjs.com/docs/how-to/performance/resolving-out-of-memory-issues/#turn-off-source-maps-generation

src/components/BlogPost/index.tsx
travjenkins, line 1: This should not be needed but going ahead and making this more strict

src/components/Grid/Card/index.tsx
travjenkins, line 1: Same as above - just making sure this stuff is 100% checked even though it should not matter.

Brenosalv (dismissed):
Lgtm

.github/workflows/deploy-preview.yml
```diff
@@ -44,8 +48,9 @@ jobs:
                   CI: true
 
             - name: Run build
-              run: yarn run build --log-pages
+              run: yarn run build:debug:network --verbose --log-pages
```
travjenkins, line 51: Somehow this seems to help :shrug: but I feel that just has to be random luck.

gatsby-config.ts
```diff
@@ -517,7 +520,7 @@ const cfg: GatsbyConfig = {
                                             placeholder: BLURRED
                                             width: 400
                                             aspectRatio: 1.7
-                                            formats: [AUTO, WEBP, AVIF]
+                                            formats: [AUTO, WEBP]
```
travjenkins, line 523: Removing support for AVIF for now as this speeds up build times
https://www.gatsbyjs.com/docs/how-to/performance/resolving-out-of-memory-issues/#disable-avif

gatsby-node.ts
```diff
@@ -689,7 +689,7 @@ export const sourceNodes: GatsbyNode['sourceNodes'] = async ({
     // console.log('sourceNodes:start');
     const pool = new pg.Pool({
         connectionString: SUPABASE_CONNECTION_STRING,
-        connectionTimeoutMillis: 8000,
+        connectionTimeoutMillis: 30 * 1000,
```
travjenkins, line 692: This is the connection timeout for `gatsby-source-pg` so matching that

Brenosalv (approved):
Lgtm

PR #767, opened 2025-04-04
Enable `cache` for production changes and strapi updates

travjenkins:

## Changes

-  Keep the cache file around for production builds
- Only keeping 1 cache per build

## Tests / Screenshots

-  adding the workflow setting for this PR to listen to strapi updates

### Shows when a connector logo is created or cached
image


gatsby-node.ts
travjenkins, line 1: One common change all over is replacing `console.log` with the proper `reporter` calls so it is properly integrated with Gatsby's build system.

gatsby-node.ts
```diff
@@ -704,93 +705,127 @@ export const sourceNodes: GatsbyNode['sourceNodes'] = async ({
...
     for (const conn of connectors.rows) {
         const usUrl = conn.logo_url?.['en-US'];
 
-        if (usUrl) {
-            const fileNode = await createRemoteFileNode({
+        // A logo is 100% required - so leave if we don't get one
+        if (!usUrl) {
+            reporter.warn(`sourceNodes:postgres:${conn.id}:missing logo`);
+            return;
+        }
```
travjenkins, line 729: To reduce nesting we just check for URL and then return.

gatsby-node.ts
```diff
@@ -704,93 +705,127 @@ export const sourceNodes: GatsbyNode['sourceNodes'] = async ({
...
-                    contentDigest: createContentDigest(fileNode),
+                    contentDigest: createContentDigest(image),
                 },
-            });
+            };
+            await createNode(nodeSettings);
+
+            // Need to cache the ENTIRE node so that all our own fields
+            //  make it back in when we touchNode
+            await cache.set(cacheKey, nodeSettings);
```
travjenkins, line 784: Cache the **entire** `createNode` settings.

This is SUPER important. The reason the logos were missing during cached builds was because we look them up with our own fields `connectorId`. So in order to make sure those fields are populated again during an incremental build we need to cache them.

package.json
```diff
@@ -107,6 +107,7 @@
     "build:debug:network": "yarn run build:plugins && NODE_DEBUG=http,http2 gatsby build",
     "build": "yarn run build:plugins && gatsby build",
     "develop": "gatsby develop",
+    "develop:debug": "gatsby develop --inspect",
```
travjenkins, line 110: I constant use this so just leaving it in here.

.github/workflows/deploy-prod.yml
```diff
@@ -70,23 +70,30 @@ jobs:
...
                   node-version: '20.x'
+            
             - name: Caching
               id: gatsby-cache-build
               uses: actions/cache@v4
               with:
                   path: |
                       node_modules
+                      .cache
+                      public
```
travjenkins, line 90: This allows GitHub Cache to keep the `gatsby` cache around

.github/workflows/deploy-preview.yml
```diff
@@ -48,9 +44,8 @@ jobs:
                   CI: true
 
             - name: Run build
-              run: yarn run build:debug:network --verbose --log-pages
+              run: yarn run build --verbose --log-pages
```
travjenkins, line 47: We should no longer need the network debugging.

gatsby-node.ts
```diff
@@ -704,93 +705,127 @@ export const sourceNodes: GatsbyNode['sourceNodes'] = async ({
...
     await pool.end();
 
-    console.log('sourceNodes:', {
-        completed: `${createdCount}/${connectors.rowCount}`,
-    });
+    reporter.info(`sourceNodes:postgres:cached:${cachedCount}`);
+    reporter.info(`sourceNodes:postgres:created:${createdCount}`);
+    reporter.info(`sourceNodes:postgres:total:${connectors.rowCount}`);
+
+    reporter.success(`sourceNodes:completed`);
```
travjenkins, line 798: Updating logging to show a break down of "stats"

Also showing the `success` flag when done.

Brenosalv (approved):
Lgtm

PR #775, opened 2025-04-10
Fix connector search debounce clear out

travjenkins:
## Changes

-   Better handling of the query string
- No longer passing around a string that was not needed

## Tests / Screenshots

### DIFF
--- a/src/components/Connectors/index.tsx
+++ b/src/components/Connectors/index.tsx
@@ -236,10 +236,9 @@ export const Connectors = ({
 
     const handleQueryChange = (evt: any) => {
         const newVal = evt.target.value;
-        setQuery(newVal);
+        setQuery(newVal ?? '');
 
         if (newVal && newVal.length > 0) {
-            console.log('fire event', newVal);
             fireTagEvent('event', 'Connector_Search', {
                 filterQuery: newVal,
             });
--- a/src/components/Connectors/index.tsx
+++ b/src/components/Connectors/index.tsx
@@ -219,9 +219,6 @@ export const Connectors = ({
         );
     }, [mappedConnectors]);
 
-    // We need to keep the input updated right away when a user types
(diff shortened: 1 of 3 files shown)


src/components/Connectors/index.tsx
```diff
@@ -250,6 +245,11 @@ export const Connectors = ({
         }
     };
 
+    const debouncedHandleQueryChange = useMemo(
+        () => debounce(handleQueryChange, 400),
+        []
+    );
```
travjenkins, line 251: We can now just debounce the entire change handler as the `SearchInput` component will not longer use the value of `query` for the input.

src/components/SearchInput/index.tsx
```diff
@@ -5,23 +5,17 @@ import {
...
-    placeholder,
-}: SearchInputProps) => {
+const SearchInput = ({ handleQueryChange, placeholder }: SearchInputProps) => {
     return (
         <div className={blogsIndexSearch}>
             <SearchIcon className={blogsIndexInputAdornment} />
             <input
                 placeholder={placeholder}
                 type="text"
-                value={query}
```
travjenkins, line 24: We do not prefill this value so no need to make it controlled.

src/components/Connectors/index.tsx
```diff
@@ -237,11 +234,9 @@ export const Connectors = ({
         showAllConnectors ? true : (res as any).type === connectorType
     );
 
-    const debounceSetQueryInput = useMemo(() => debounce(setQuery, 400), []);
     const handleQueryChange = (evt: any) => {
         const newVal = evt.target.value;
-        setQueryInput(newVal);
-        debounceSetQueryInput(queryInput);
+        setQuery(newVal ?? '');
```
travjenkins, line 239: Being super safe.

Brenosalv (approved):
Lgtm

Issue #749, opened 2025-03-27
Embedd Case Study Card


danthelion:
Airbyte vs Estuary -> https://estuary.dev/success-stories/headset

danthelion:
mockup idea: 

[short text from case study] + logo

<img width="1566" alt="Image" src="https://github.com/user-attachments/assets/d1ab1bdf-ad44-42e2-9068-e20b75293fba" />

Brenosalv:
@danthelion Should we have this on all VS pages (Vendor X vs Vendor Y) or only the ones with Estuary Flow (Vendor X vs Estuary Flow)?

Brenosalv:
@danthelion Is it a new section with heading or just a part of the introduction section? If it's a section, how should be the heading? Maybe "Success Story" or "Headset Success Story" or "Estuary Flow Success Story".

Not sure if it fits well between the intro and comparison matrix sections. It fits better down below on the Estuary Flow details section after the pros, cons and pricing. Thoughts?

<img width="481" alt="Image" src="https://github.com/user-attachments/assets/3acbb842-851e-4670-8b7e-919fa3a899cb" />

danthelion:
I didn't want to push it so far down. @SourabhEstuary , any thoughts on where we could place a card in the VS pages for success stories?

Regarding the header, we could do something like:

Header: Success Story
Subheader: "Check out how Headset elevated their data stack by switching from Airbyte to Estuary"

SourabhEstuary:
If it's only being added to "Estuary vs Competitor," then placing it higher, right after the introduction, is fine.

danthelion:
updated scope: 

1. re-usable embeddable case study card that can be added to integration pages, comparison pages, blog posts.
2. card has text from case study model short description
3. card should have button [read success story] which when clicked navigates to success story
4. Header and text comes from Strapi.

PR #779, opened 2025-04-14
Add success story banner to the comparison page

Brenosalv:
#749 

## Changes

-   Create component for the success story banner and use headset as default for now;
-   Add it to the comparison page;
-   For now we will use the banner only in the estuary vs airbyte comparison page;
-   Button is opening the headset success story page in a new tab.

## Tests / Screenshots

-   Desktop:
image

-   Mobile:
image

### DIFF
--- a/src/components/SuccessStoryBanner/styles.module.less
+++ b/src/components/SuccessStoryBanner/styles.module.less
@@ -1,0 +1,59 @@
+.container {
+  display: flex;
+  align-items: center;
+  gap: 24px;
+  background-color: #FFA80050;
+  border: 1px solid #FFA800CC;
+  border-radius: 12px;
+  padding: 24px;
+
+  img {
+    width: 100%;
+    height: auto;
+  }
+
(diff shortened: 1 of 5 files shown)


src/components/SuccessStoryBanner/styles.module.less
```diff
@@ -0,0 +1,59 @@
+.container {
+  display: flex;
+  align-items: center;
+  gap: 24px;
+  background-color: #FFA80050;
+  border: 1px solid #FFA800CC;
```
travjenkins, line 6: What is driving having these colors "hardcoded" vs reading from a css var? Are they one offs?

Brenosalv: It really makes sense to create variables for these colors since they are used in both release banner from the homepage and success story banner. Thanks for the suggestion.

travjenkins (approved):
lgtm - did not manually test

Issue #723, opened 2025-03-03
Implement RSS Feed for Blog (Only Blogs)


SourabhEstuary:
Create RSS Feed for the Blog Section

- Ensure the feed includes only published blog posts, not other pages.
- Each <item> in the feed should have: Title (<title>), URL (<link>) – should follow /blog/{slug} format, Description (<description>), Published Date (<pubDate>), Updated Date (<lastBuildDate>)
- Ensure the feed automatically updates when: A new blog post is published, An existing blog post is updated.

Rss Feed example 

![Image](https://github.com/user-attachments/assets/ce70e213-ac92-42fa-a571-cb23a1a6c8f1)

travjenkins:
This has been added and removed in the past. No clue why it was removed since it was only there during the initial commit - so solid chance it never ran in production.

![Image](https://github.com/user-attachments/assets/66aad5a2-af72-4f30-8fc3-3d41f125d9e4)

So we should just uncomment that section and get it configured

SourabhEstuary:
Can we check when this was commented out and removed? I'll analyze whether its removal had any positive or negative impact on SEO performance.

travjenkins:
It was added and removed before the site was even live.

SourabhEstuary:
If this doesn’t take much time, can we prioritize it? I’d like to use the feed to submit our blog to some distribution sites.

Brenosalv:
@SourabhEstuary Should we limit the number of blog posts or you need all of them in the feed?

SourabhEstuary:
We need all blog posts, both existing and future, included in the RSS feed.  Feed should update automatically whenever a new blog is published without requiring any manual intervention.

PR #790, opened 2025-04-18
Uncomment rss feed plugin and configure it

Brenosalv:
#723

## Changes

-   Uncomment rss feed plugin and configure it.

## Tests / Screenshots

-   The RSS feed XML is generated only in production. Tested in preview:

https://estuary-marketing--pr790-brenosalv-feature-72-huxbi2fa.web.app/blog/rss.xml
image

RSS feed is valid:
image

### DIFF
--- a/gatsby-config.ts
+++ b/gatsby-config.ts
@@ -717,10 +717,11 @@ const cfg: GatsbyConfig = {
               `,
                 feeds: [
                     {
-                        feed_url: ({ query: { site } }) =>
-                            `${site.siteMetadata.siteUrl}/blog/rss.xml`,
-                        site_url: ({ query: { site } }) =>
-                            site.siteMetadata.siteUrl,
+                        custom_namespaces: {
+                            atom: 'http://www.w3.org/2005/Atom',
+                        },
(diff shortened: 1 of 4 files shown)


gatsby-config.ts
```diff
@@ -13,6 +13,8 @@ import { SUPABASE_CONNECTION_STRING } from './config';
 // Disable multiple prepared statements because pgbouncer doesn't like 'em very much
 process.env.POSTGRAPHILE_PREPARED_STATEMENT_CACHE_SIZE = '1';
 
+const BLOG_PATH_REGEX = '^/blog/';
```
travjenkins, line 16: I think this is incorrect as it is now not including the link on any page I can find.

travjenkins (changes requested):
Looks to have an issue with adding the link

gatsby-config.ts
Brenosalv: It happened because probably you tested locally and the plugin only generates the xml file when running in production mode. You can test the preview.

Brenosalv: Got it, I was testing only going directly to the blog page.

Brenosalv:
Validation result in production:
image

Issue #554, opened 2024-11-01
Expandable FAQ Section with FAQ Schema for Blog Posts


SourabhEstuary:
Please add an optional expandable FAQ section at the end of our blogs, similar to the one shown in the attached screenshot. This section should only be visible when we have specific FAQs to include; otherwise, it should be hidden entirely.

Also, please ensure that once this section is added, the FAQ schema is implemented for it.

Let me know if you need any further details. Thank you!
![image](https://github.com/user-attachments/assets/a9080f0d-0686-4c62-96a3-ec580c75316e)

travjenkins:
@danthelion - we need to work out a lot more details / discussion.

What content goes in here?
What does the shape of the content look like?
Do all blogs have this content?
How are they tied to blogs?

We need to work out way more details on this.

SourabhEstuary:
**What content goes here?**
This section is optional. We'll add 3-4 FAQs at the end of certain blogs to provide additional helpful information. For reference, see an example of currently added FAQs: - https://estuary.dev/netsuite-to-sql-server/ . 

**What should the content look like?**
Each FAQ answer will typically be around 50 words, though some may extend to about 70-75 words.

**Will all blogs have FAQs?**
No, FAQs are for some priority blogs, primarily for source-to-destination blogs. If there’s no specific FAQ content, this section should remain hidden. This should be managed by strapi.

**Where will the FAQs appear in the blog?**
FAQs will be placed at the end of the blog, just above our final CTA section. 

*A dedicated FAQ section can improve the reader experience and make the content more helpful. It also increases the chances of boosting SEO performance.

Brenosalv:
We can add optional FAQ fields to the blog posts in Strapi and reuse the FAQ UI from the pricing page.

travjenkins:
Are you wanting to share these or will they always be fully unique for the blog post?

danthelion:
They'll be unique to the blog on every occasion

travjenkins:
Okay cool - yeah just having an array of these on blogs sounds good.

SourabhEstuary:
Hey team, can we prioritize expandable FAQ section? Let me know if there are any blockers...

travjenkins:
Do y'all have the content ready?

SourabhEstuary:
Yes, we already have FAQs added to many published blogs. Once a manageable FAQ section is available in Strapi, we’ll start adding FAQs to new blogs and gradually update existing ones over time. Thanks!

[...]

PR #792, opened 2025-04-20
Add expandable FAQ section with FAQ schema for blog posts

Brenosalv:
#554

## Changes

-   Add faq to blog post query;
-   Create a component FaqAccordions to reuse the FAQ sections code;
-   Add FAQ section to the blog post right above the last CTA banner;
-   Add FAQ schema to the Head of the blog post page.

## Tests / Screenshots

-   Blog post FAQs:
image

-    Blog post FAQ schema:
image

### DIFF
--- a/src/components/BlogPost/Faqs/index.tsx
+++ b/src/components/BlogPost/Faqs/index.tsx
@@ -2,7 +2,7 @@ import FaqAccordions from '../../Faq/FaqAccordions';
 import { container } from './styles.module.less';
 
 const Faqs = ({ faqs }) => (
-    <div id="faq" className={container}>
+    <div id="blog-post-hardcoded-faq" className={container}>
         <h2>FAQ</h2>
         <FaqAccordions faqs={faqs} />
     </div>
--- a/src/components/BlogPost/Faqs/index.tsx
+++ b/src/components/BlogPost/Faqs/index.tsx
@@ -3,15 +3,13 @@ import Container from '../../Container';
 import FaqAccordions from '../../Faq/FaqAccordions';
 import { sectionTitle } from './styles.module.less';
 
-const Faqs = ({ faqs }) => {
-    return (
-        <section className={defaultWrapperDark}>
-            <Container isVertical>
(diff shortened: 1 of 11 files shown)


src/templates/blog-post/index.tsx
travjenkins, line 1: Could there be an issue if we are always adding this to the schema even when there is not a FAQ on the blog post?

Brenosalv: Not sure if it would be an issue but it really doesn't make sense to have FAQ schema for the blog posts that don't have FAQ. So I will add a condition.

src/components/BlogPost/index.tsx
```diff
@@ -189,7 +195,12 @@ const BlogPost = ({
                                 slug: post.slug,
                             }}
                             tableOfContents={
-                                post.body.data.childHtmlRehype.tableOfContents
+                                post?.faq?.length > 0
+                                    ? [
+                                          ...bodyToc,
+                                          { id: 'faq', heading: 'FAQ' },
```
travjenkins, line 201: can we update the `id` to be super unique just to be safe. Something like `hardcoded-faq`?

src/components/BlogPost/Faqs/index.tsx
```diff
@@ -0,0 +1,11 @@
+import FaqAccordions from '../../Faq/FaqAccordions';
+import { container } from './styles.module.less';
+
+const Faqs = ({ faqs }) => (
+    <div id="faq" className={container}>
```
travjenkins, line 5: Like down below... I think we should make this something like `hardcoded-faq`?

travjenkins (changes requested):
1 tiny change

src/components/BlogPost/index.tsx
Brenosalv: I added `blog-post-hardcoded-faq` since it's only on the blog post pages.

travjenkins (approved):
lgtm

Issue #784, opened 2025-04-16
New Testimonial Layout Style


danthelion:
design: https://www.figma.com/design/qhXFvOJPqLgGsr9lsAPYbn/Estuary--Website?t=MD1PyYRxo3uHDy9W-0

travjenkins:
https://company.g2.com/brand-resources

Brenosalv:
Add an optional related success story field in Strapi for Testimonial.

danthelion:
- g2 part: just the g2 logo and the stars

PR #809, opened 2025-04-28
New testimonial layout style

Brenosalv:
#784

## Changes

-   Update testimonial to new layout and content;
-   Adjust the testimonial section from integration page.

## Tests / Screenshots

-   Desktop:
image

- Mobile:
image


src/components/EtlToolsPage/ListOfComparisons/index.tsx
```diff
@@ -89,14 +89,18 @@ const ListOfComparisons = ({ estuaryVendor, vendors }: SectionThreeProps) => {
...
+                                            <GatsbyImage
+                                                image={
+                                                    logo.localFile
+                                                        .childImageSharp
+                                                        .gatsbyImageData
+                                                }
+                                                alt={`${name} Logo`}
+                                                className={itemImage}
+                                            />
+                                        ) : null}
```
Brenosalv, line 103: It's an off topic. I visited this component and there was an error. We needed to check if `logo.localFile.childImageSharp` is not undefined. Below is the type imported from `shared.ts`.

```
interface ImageType {
    childImageSharp?: {
        gatsbyImageData: IGatsbyImageData;
    };
}

logo: {
    localFile: ImageType;
};
```

src/components/Homepage/WhatPeopleAreSaying/index.tsx
```diff
@@ -1,16 +1,41 @@
...
+                        <div className={g2Wrapper}>
+                            <G2LogoIcon width={64} height={64} />
+                            <div className={starsWrapper}>
+                                <FiveStarsIcon width={116} height={20} />
+                                <span>4.8/5</span>
+                            </div>
+                        </div>
+                        <p>
+                            Hear how Estuary is making a difference for
+                            businesses like yours.
```
Brenosalv, line 32: @danthelion Maybe we should replace "Hear" with "Read"?

src/components/TestimonialsCarousel/TestimonialAvatar/index.tsx
```diff
@@ -7,9 +7,13 @@ import {
     isSeattleDataGuyLogo,
 } from './styles.module.less';
 
-const imgSize = 110;
+const imgSize = 67;
 
 const TestimonialAvatar = ({ logo, name }) => {
+    if (!logo?.localFile) {
+        return null;
+    }
```
Brenosalv, line 15: In case the testimonial does not have the author avatar, it returns `null`.

Brenosalv:
Testimonials with big quotes - I will remove the little quotes once we are ready to merge:
image

src/components/Carousel/index.tsx
```diff
@@ -56,14 +63,28 @@ const Carousel = ({
...
                 container,
-                hasMultipleItemsSlide ? threeItemsSlide : null
+                hasMultipleItemsSlide ? threeItemsSlide : null,
+                rest.className
             )}
+            style={
+                Object.keys(containerStyles).length
+                    ? containerStyles
+                    : undefined
+            }
```
travjenkins, line 85: Since you defined the object - it is safe to just always set this.

src/components/EtlToolsPage/ListOfComparisons/index.tsx
travjenkins: We need to get away from the approach. We control the application and the data we are fetching. We should be checking that the data is there while fetching and if it is not then the build should fail. We do not want the site being deployed in a state that we do not expect.

src/components/TestimonialsCarousel/TestimonialAvatar/index.tsx
```diff
@@ -1,38 +1,69 @@
...
+                    [
+                        'Resend',
+                        'Forward',
+                        'Headset',
+                        'Recart',
+                        'Revunit',
+                    ].includes(name)
+                    ? wideLogo
+                    : null,
+                ['Flashpack', 'Pompato'].includes(name) ? notRoundedLogo : null
```
travjenkins, line 49: This is okay for now - but this should not be handled this way. These are options we should be setting in strapi.

src/components/TestimonialsCarousel/index.tsx
```diff
@@ -46,28 +71,65 @@ const TestimonialsCarousel = ({
...
+                            <InternalLink
+                                href={`/success-stories/${relatedSuccessStory.slug}`}
+                                target="_blank"
+                            >
+                                Read the Success Story
+                                <ArrowRightIcon />
+                            </InternalLink>
+                        ) : null}
+                    </div>
+                )
```
travjenkins, line 131: This is big enough I think it should probably be worked into a component.

travjenkins (changes requested):
just two changes

src/components/TestimonialsCarousel/TestimonialAvatar/index.tsx
Brenosalv: Yeah, I also don't like this but the logos don't have a pattern and it was not possible to make them good with a unique style. Some logos have symbol/emblem and text, some only emblem and some only text. To apply the same style to all we would need them to have the same size like all having both emblem and text. I'm not sure how much we can manipulate the logos.

Brenosalv: Or you mean having a different field for each type of logo in Strapi?

travjenkins: Yeah - potentially a flag.

We could also start breaking up the `company` and `mark` portion into two different images.
image

travjenkins (dismissed):
lgtm

src/components/EtlToolsPage/ListOfComparisons/index.tsx
Brenosalv: Makes sense. I will make it required.

src/components/TestimonialsCarousel/TestimonialAvatar/index.tsx
Brenosalv: I worked more on the CSS and figured out a way to have a unique CSS for all logos. Also resized and removed background of some logos like OneCommerce, Minima Global and Seattle Data Guy.

image

Brenosalv:
Already with no little quotes and with big quotes:

image

travjenkins (approved):
lgtm

Issue #689, opened 2025-02-13
startup program page


danthelion:
design: https://www.figma.com/design/qhXFvOJPqLgGsr9lsAPYbn/Estuary%3A-Website?node-id=11254-24349&t=CHNfNShmEEHPE2W2-0

PR #817, opened 2025-05-05
Startup program page

Brenosalv:
#689

## Changes

-   Add sections according to Figma. Check Figma in the issue description.

## Tests / Screenshots

-   Tested locally and looks good.

## Preview with slug

- https://estuary-marketing--pr817-brenosalv-feature-68-jjlca6e2.web.app/startup


src/components/StartupPage/WhatFoundersAreSaying/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+                        “We needed a platform that allowed us to give our
+                        customers more ways to analyze and visualize their
+                        social media performance vs their competitors, and scale
+                        with our rapid growth. With Estuary we were able to get
+                        enterprise-grade social analytics up in a few days with
+                        second-level latency and performance.”
+                    </p>
+                    <strong>Shane Iseminger</strong>
+                    <span>CTO, SocialHP</span>
+                </div>
```
travjenkins, line 39: For stuff like quotes I think we should look at using semantic tagging. Can you see if we could switch this over to `blockquote`

travjenkins (changes requested):
Mentioned some small issues with our chat and one callout here

src/components/StartupPage/WhatFoundersAreSaying/index.tsx
Brenosalv: Yes, we can use `blockquote`. I think it won't impact a lot since we won't use the default style of blockquote like the left border and paddings around the text, italic, etc. However, at least for accessibility it might help. So I will add it besides the element `cite` replacing the `span`.

travjenkins (dismissed):
lgtm

travjenkins (dismissed):
lgtm

travjenkins (approved):
Lgtm

Issue #760, opened 2025-04-02
Migrate from `dayjs` to `luxon`


Brenosalv:
_Originally posted by @Brenosalv in https://github.com/estuary/marketing-site/pull/748#discussion_r2025334913_

PR #819, opened 2025-05-07
Remove `dayjs` dependency

Brenosalv:
#760

## Changes

-   Remove dayjs dependency because it's not being used anywhere in fact. We already are formatting dates directly in the GraphQL queries.
-   So no need for replacing `dayjs` with `luxon`.

## Tests / Screenshots

-   Tested the blog posts, company update posts and success stories cards and the dates still look good as before:

image
image
image


travjenkins (approved):
lgtm

Issue #828, opened 2025-05-14
Data Subject Access Request Material


skord:
# Data Subject Access Request Material
Add dedicated **/data-requests** page and footer link for GDPR/DSAR handling

---

## Why we’re doing this
Regulators (and privacy-savvy customers) must be able to locate a clear, single location for exercising data-protection rights. Hosting the actual request form on Google Forms keeps it outside the site’s marketing/analytics stack, preventing accidental CRM enrichment.

---

## Scope of work  

1. **New page:** `/data-requests`  
   * **URL:** `/data-requests`   
   * **Content (all that’s needed):**

     > To exercise your data-protection rights, please complete our <a href="https://forms.gle/uhKXUE7AnJsM3fc96" target="_blank" rel="noopener">Data Request Form</a> or email <a href="mailto:privacy@estuary.dev">privacy@estuary.dev</a>.<br><br>
     > Please do not send sensitive documents (e.g., passport scans or other identification) via email. If verification is required, we will provide a secure upload link.

2. **Footer link**  
   * Add **“Data Requests”** next to existing Privacy-/ToS links.  
   * Destination: `/data-requests`.

3. **Analytics exclusion**  
   * Confirm `/data-requests` is **excluded** from marketing events, pixel fires, and lead-capture flows.

---

## Acceptance criteria  

- [x] `/data-requests` page renders in production with the exact wording above.  
- [x] Footer link visible on all pages where footer is rendered
- [x] Page is excluded from analytics / marketing automation (GA4, HubSpot, etc.).  
- [x] PR reviewed by Security lead.

travjenkins:
This has been merged and is in the process of being deployed.

PR #841, opened 2025-05-21
Travjenkins/compliance/add data request page

travjenkins:
## Changes

https://github.com/estuary/marketing-site/issues/828
-   changes...

## Tests / Screenshots

### DIFF
--- a/src/pages/data-requests.tsx
+++ b/src/pages/data-requests.tsx
@@ -1,0 +1,44 @@
+import Layout from '../components/Layout';
+import Seo from '../components/seo';
+import { globalMaxWidth } from '../globalStyles/sections.module.less';
+
+// - COMPLIANCE WARNING -
+// DO NOT INCLUDE TRACKING ON THIS PAGE
+// file name checked in `gatsby-ssr`
+const DataRequestPage = () => {
+    return (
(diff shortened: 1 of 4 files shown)


src/pages/data-requests.tsx
```diff
@@ -0,0 +1,46 @@
...
+                    Please do not send sensitive documents (e.g., passport scans
+                    or other identification) via email. If verification is
+                    required, we will provide a secure upload link.
+                </p>
+            </div>
+        </Layout>
+    );
+};
+
+export const Head = () => <Seo title="Estuary data-protection rights" />;
```
travjenkins, line 44: @skord - any specific title we need to have for this page?

skord: **Data Subject Access Request**

Thanks for asking. It’s one of those things where the closer we get it to the GDPR language (in this case, exact) the better due to the fact that we’re not translating it into other languages.

skord: It’s fine if it’s indexed, but on a related note, we cannot have any tracking on this page.

skord (approved):
I am very happy with this. Thank you so much. LGTM

PR #845, opened 2025-05-25
Add contact us form when connectors list is empty

Brenosalv:
#840 

## Changes

-   Add contact us form when there is no connectors found.

## Tests / Screenshots

- Desktop
image

- Mobile
image

### DIFF
--- a/src/components/Connectors/index.tsx
+++ b/src/components/Connectors/index.tsx
@@ -32,8 +32,6 @@ import {
     connectorCard,
     connectorBottomButton,
     contactUsForm,
-    contactUsFormWrapper,
-    hideContactUsForm,
 } from './styles.module.less';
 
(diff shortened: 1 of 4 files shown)


travjenkins (dismissed):
lgtm - the only issue might be if a user is searching a lot and constantly switching between results and no results. As this will trigger the loading several times.

Probably an edge case so I think we are okay for now.

Brenosalv:
@travjenkins yeah, I will take a look if we can do something about this edge case and will update the "No results found" text as requested by Dani. I let you know.

Brenosalv:
@travjenkins For solving the edge case issue, we can keep the form mounted and hide it with CSS when we have search results. I added a new commit for this.

travjenkins (approved):
lgtm

Issue #834, opened 2025-05-16
Add search for success stories


danthelion:
like blog page

Brenosalv:
Remove this section:

<img width="1463" alt="Image" src="https://github.com/user-attachments/assets/17cd69c7-df1c-4354-b275-f2abf9444c73" />

Brenosalv:
Add related tags field for success stories in Strapi.

PR #848, opened 2025-05-28
Add search for success stories

Brenosalv:
#834

## Changes

-   Add search for success stories on success stories page;
-   Drop "Featured customers" section from the same page;
-   Add a "No success stories found" when search results is empty.

## Tests / Screenshots

- Desktop
image

- Desktop (searching):
image

- Mobile
image

### DIFF
--- a/gatsby-config.ts
+++ b/gatsby-config.ts
@@ -180,6 +180,9 @@ const strapiConfig = {
                             },
                         },
                     },
+                    tags: {
+                        populate: '*',
+                    },
                 },
             },
         },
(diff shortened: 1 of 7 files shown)


shared.ts
```diff
@@ -279,3 +280,64 @@ export const getIntegrationSlug = (
     connectorName1?: string,
     connectorName2?: string
 ) => `/integrations/${connectorName1}-to-${connectorName2}`;
+
+export function buildLunrQuery(query: string) {
```
travjenkins, line 284: If we are going to share then this function needs to be named a bit better. That way if we add a 3rd search someone does not try to use this on as the "default" since this is closely tied to blogs/use cases.

gatsby-config.ts
```diff
@@ -696,6 +699,93 @@ const cfg: GatsbyConfig = {
...
+            options: {
+                // A unique name for the search index. This should be descriptive of
+                // what the index contains. This is required.
+                name: 'cases',
+
+                // Set the search engine to create the index. This is required.
+                // The following engines are supported: flexsearch, lunr
+                engine: 'lunr',
+
+                // GraphQL query used to fetch all data for the search index. This is required.
```
travjenkins, line 713: I don't think we need to duplicate all the comments from the original.

gatsby-config.ts
```diff
@@ -696,6 +699,93 @@ const cfg: GatsbyConfig = {
...
+
+                // List of keys to index. The values of the keys are taken from the
+                // normalizer function below.
+                // Default: all fields
+                index: ['slug', 'title', 'searchable_tags'],
+
+                // List of keys to store and make available in your UI. The values of
+                // the keys are taken from the normalizer function below.
+                // Default: all fields
+                store: ['id', 'title', 'slug', 'tags', 'description', 'hero'],
```
travjenkins, line 757: Since these are getting shared the default settings should probably be stored in a variable close to the new function. That way we know to keep them in sync.

gatsby-config.ts
```diff
@@ -696,6 +699,93 @@ const cfg: GatsbyConfig = {
...
+
+                // Function used to map the result from the GraphQL query. This should
+                // return an array of items to index in the form of flat objects
+                // containing properties to index. The objects must contain the `ref`
+                // field above (default: 'id'). This is required.
```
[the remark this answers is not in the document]
Brenosalv: Okay, I removed the general comments from the connector and success stories plugins, but kept them for the blog to use as a reference. I also left the specific comments (there was only one for connectors search plugin) in place.

gatsby-config.ts
Brenosalv: The connector plugin was repeating the comments from blog plugin, so I thought it was a pattern to make it clear when creating a new search plugin. But it makes sense not to repeat it, we can always use the original (blog) as the reference when creating a new one.

gatsby-config.ts
```diff
@@ -618,16 +617,10 @@ const cfg: GatsbyConfig = {
...
             resolve: 'gatsby-plugin-local-search',
             options: {
-                // A unique name for the search index. This should be descriptive of
-                // what the index contains. This is required.
                 name: 'connectors',
 
-                // Set the search engine to create the index. This is required.
-                // The following engines are supported: flexsearch, lunr
-                engine: 'lunr',
+                engine: gatsbyPluginLocalSearchSettings.engine,
```
travjenkins, line 622: This is okay... in the future it would be easier to just spread up above:

```
options: {
    ...gatsbyPluginLocalSearchSettings,

```

Issue #780, opened 2025-04-15
Deprecated Integration Pages


SourabhEstuary:
Please add a **noindex tag** to all deprecated connector integration pages. Just make sure it’s only applied to those pages, and not to any active or supported connector integration pages.

If we're planning to remove these pages entirely from the website, please set up a 301 redirect, ideally to the most relevant active connector page. If no relevant page exists, redirect to the homepage.

SourabhEstuary:
these are deprecated connector landing pages in the website. Any integration pages generated using these will also be considered deprecated and should be marked with noindex


- https://estuary.dev/source/shopify/
- https://estuary.dev/source/jira/
- https://estuary.dev/source/mixpanel/
- https://estuary.dev/source/salesforce/
- https://estuary.dev/source/stripe/
- https://estuary.dev/source/intercom/
- https://estuary.dev/source/zendesk-support/
- https://estuary.dev/source/google-sheets/
- https://estuary.dev/source/salesforce-next/

Brenosalv:
@danthelion Please let me know if these deprecated connector pages and their integration pages should be removed or we just keep them live without indexing them.

travjenkins:
There are several more deprecated connectors that the team has updated:

source-redshift - had (OLD) in the title & no active captures
source-chargebee - we have a native version now
source-google-analytics-data-api - we have a native version now
source-jira-legacy - we have a native version now
source-linkedin-ads - this is the old ATF connector, we have an imported version
source-oracle-flashback - no longer supported
materialize-rockset - no longer supported
source-hubspot -

PR #855, opened 2025-06-03
More tweaks to the deprecated connectors

Brenosalv:
#780

## Changes

-   Drop deprecated connectors from list on `/integrations` page;
-   Add missing redirects.

## Tests / Screenshots

-   Tested locally and looks as expected.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -95,6 +95,8 @@ export const normalizeConnector = (
                 `Error:connector:${id}:missing redirect for deprecated connector (title contains “deprecated” but no redirect entry found)`
             );
         }
+
+        return undefined;
     }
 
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -95,6 +95,8 @@ export const normalizeConnector = (
                 `Error:connector:${id}:missing redirect for deprecated connector (title contains “deprecated” but no redirect entry found)`
             );
         }
+
+        return undefined;
```
travjenkins, line 99: Can you please move the entire check up a bit? We should catch this quick as we no longer care if there isn't a logo, image name, etc.

image

travjenkins (approved):
lgtm

Issue #852, opened 2025-06-03
Add all fields to success stories search index


Brenosalv:
## Description

Currently we index only the title, slug and tags on success stories search feature. We need to index all the fields.

## Work required

-   Add the other fields to index;
-   Adapt the rich text to the search settings.

PR #867, opened 2025-06-10
Add other fields to success stories search index

Brenosalv:
#852

## Changes

-   Add fields in the plugin settings;
-   Update the `searchIndex` function with the new fields;

## Tests / Screenshots

Some searching examples below:

image

image

image


Brenosalv:
When I search for “Airbyte,” only the “Headset” success story shows up - even though there should be others. I’m not sure why the other stories aren’t being matched. Some success stories like “Shippit” contains that term in its rich-text fields.

Brenosalv:
Some tests:

- Blog search looks good:
image

- Searching blog post by author name:
image

- Tested the connectors and they are the same.

Brenosalv:
We now handle typos in searches.

travjenkins (approved):
lgtm

Issue #847, opened 2025-05-27
Add FAQs schema


SourabhEstuary:
Add FAQs schema in Integration pages and Pricing page FAQ section.

PR #873, opened 2025-06-12
Add FAQ schemas to pricing and integration pages

Brenosalv:
#847

## Changes

-   Add FAQ schemas to pricing and integration pages.

## Tests / Screenshots

-   Pricing page FAQs schema:
image

-   Integration page FAQs schema:
image


SourabhEstuary:
> Do we normally not just include the marks we want in SEO content? image

Yes..Characters like &#x27; should definitely be cleaned up so the text reads normally.

SourabhEstuary:
> I will leave it up to @SourabhEstuary - but I am not totally sure we should be putting this content in exactly as it shows up in the site.
> 
> Things like the `headers` make more sense visually but seems less important when using text in an answer. image
> 
> I kind of think we should break this kind of stuff out of the answer as it seems all the same details are in the paragraph below it. image

As for the <h4> and <h5> headers (e.g. “STREAMING ETL”), those make sense visually on the page, but not in the schema's text field. They should be excluded from the structured data, and only plain paragraph content should be used.

Brenosalv:
Now we are considering only the paragraphs and handling the HTML entities. Tested locally and looks good:

- Pricing page FAQs schema:
image

- Integration page FAQs schema:
image

Issue #866, opened 2025-06-09
New tech solutions page for Snowflake


aeluce:
## Description
We'd like a new solutions page showcasing Snowflake that uses a different layout from the current Strapi structure. This may require a handcrafted page with hardcoded content. If, however, the Strapi structure can be modified to accommodate the new design, existing pages should still continue to work, and it would be nice if it allowed for additional flexibility, like picking and choosing or reordering sections.

Features of note:

* Video: Embedded video instead of a hero graphic. May be focused entirely on the Snowflake connector in the future, but we can use this one for now: https://youtu.be/g1OFwvEmbkw
* Resource carousel: Showcase relevant resources. Should be able to handle/pull correct thumbnail and description from both case studies and blog posts.
* Charts: Some sections include charts between blocks of text. It might be nice, but not necessary, if these charts were programmatic so that if data changes (like if Snowpipe Streaming becomes even faster), we'd be able to easily update the chart values. Otherwise, static images are fine.
* CTA buttons: It'd be great to have CTA buttons at the bottom of the page so that if someone reads through all the information and wants to try us out, they don't have to scroll back up.

## Work required
-   [x] Create new section fields in Strapi admin to use in the new solution page;
-   [x] Configure and fetch new fields from Strapi on the marketing site;
-   [x] Implement the UI for the new solution page sections on the marketing site.

## Figma / Designs

Based on this design: https://www.figma.com/design/qhXFvOJPqLgGsr9lsAPYbn/Estuary--Website?node-id=14363-38400&t=3capg0Ay5W5ryv9t-1

Which is based on this content: https://docs.google.com/document/d/1v03e18SnJ_BVK8ezvqTDgD_ovVRW_7k--y8f3PKZHXA/edit?tab=t.l6sora8chakg

PR #884, opened 2025-06-26
New tech solutions page for snowflake

Brenosalv:
#866

## Changes

-   Configure and fetch new solution template fields from Strapi;
-   Implement new section for the new tech solution page: Snowflake.

## Tests / Screenshots

-   Tested locally and looks good. Please use the link in the solutions menu.
-   Tested the preview and looks good:
image


src/components/Solutions/List/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+                {items.map((item, index) => (
+                    <li key={`solution-section-list-item-${index + 1}`}>
+                        <div className={iconWrapper}>
+                            <Checkmark
+                                color="var(--green)"
+                                width={iconSize}
+                                height={iconSize}
+                            />
+                        </div>
+                        <ProcessedPost body={item.text.data.text} />
```
Brenosalv, line 27: Maybe now makes more sense to rename this component to `ProcessedHtml` since it's not being used only for blog posts anymore?

src/templates/solutions/shared.ts
```diff
@@ -38,26 +41,120 @@ export interface KeyFeaturesSectionContent extends SectionText {
...
+        };
         solution: TemplatePageProps;
     };
 }
+
+const snowflakeSolutionSlugInStrapi =
+    '/technology/real-time-snowflake-streaming';
+
+export const checkSnowflakeSolution = (slug: string) =>
+    slug.includes(snowflakeSolutionSlugInStrapi);
```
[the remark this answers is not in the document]
travjenkins: I think this is okay for a temporary solution.

Could we maybe detect this based on the data we have that is unique? Like possibly checking if the data contains `graphicSections`, `carouselSection`, `oneTitleThreeCards`, and `buttonsSection`
image

src/components/Homepage/BuildInMinutes/index.tsx
```diff
@@ -19,7 +19,7 @@ const BuildInMinutes = () => {
                     </p>
                 </div>
                 <div>
-                    <FlowDemoVideo />
+                    <FlowDemoVideo pageId="build-in-times-section/homepage" />
```
travjenkins, line 22: Unless this `pageId` already existed - I think we should change this. It feels a bit off to me. Maybe just following the component name works better? `build-in-minutes-section/homepage`

src/components/Solutions/SectionTitle/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+                </>
+            ) : (
+                <>
+                    <span>{sectionTitle?.highlightedText}</span>{' '}
+                    {sectionTitle?.normalText}
+                </>
+            )}
+        </h2>
+    );
+};
```
travjenkins, line 23: Could we solve this instead with CSS `flex-direction` and using a `reverse`?

src/components/Solutions/Sections/CarouselSection/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+            ...successStories.map(({ slug, ...rest }) => ({
+                ...rest,
+                slug: `/success-stories/${slug}/`,
+                type: cardType.successStory,
+            })),
+            ...blogPosts.map(({ slug, ...rest }) => ({
+                ...rest,
+                slug: `/blog/${slug}/`,
+                type: cardType.article,
+            })),
```
travjenkins, line 35: Just double triple checking.... it is safe to use `/blog/...` cause either these will link to new articles OR to old ones that have redirects handled.

gatsby-config.ts
```diff
@@ -15,9 +15,15 @@ process.env.POSTGRAPHILE_PREPARED_STATEMENT_CACHE_SIZE = '1';
...
+
 const strapiConfig = {
-    apiURL: process.env.STRAPI_API_URL,
-    accessToken: process.env.STRAPI_TOKEN,
+    apiURL: isProd
+        ? process.env.STRAPI_API_URL
+        : process.env.STRAPI_API_URL_DEV,
+    accessToken: isProd
+        ? process.env.STRAPI_TOKEN
+        : process.env.STRAPI_TOKEN_DEV,
```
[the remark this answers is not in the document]
Brenosalv: I’ve added these variables to the `readme`, but I noticed we don’t have a `.env.example` file and the `readme` doesn’t list the other environment variables. How about creating a `.env.example` and updating the `readme` to include all required variable names?

src/components/Homepage/BuildInMinutes/index.tsx
Brenosalv: Wondering why I used "times" instead of "minutes" 🤔.

I will use the component name. By the way, the component name is based on the old video section with different section title. Would you like me to start updating the component names once the section titles change? For this case it would become 

- Component name: "DeliverValueInMinutes";
- `pageId`: `deliver-value-in-minutes-section/homepage`.

image

src/components/Solutions/SectionTitle/index.tsx
Brenosalv: Done

Brenosalv: No, it's not possible to handle this with CSS only.

gatsby-config.ts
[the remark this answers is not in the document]
travjenkins: Yeah that works for me. As always - just make sure you set the values to something like `replace me` or something.

src/components/Homepage/BuildInMinutes/index.tsx
travjenkins: Nope - I generally try to get component names to be more related to what they are doing so there isn't an issue when the content, title, etc. changes.

travjenkins (dismissed):
Overall this looks good to me.

src/templates/solutions/shared.ts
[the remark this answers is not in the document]
Brenosalv: Actually it's scalable, I worked on this change and works well. Once we have a new solution page, we can add a new condition to consider only the filled sections in Strapi. You can take a look at the new code.

gatsby-config.ts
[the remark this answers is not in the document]
Brenosalv: Updated

PR #905, opened 2025-07-18
revert openid rewrites

mdibaiee:

## Changes

-   changes...

## Tests / Screenshots


firebase.json
```diff
@@ -4526,6 +4526,15 @@
         "ignore": ["firebase.json", "**/.*", "**/node_modules/**"],
         "cleanUrls": true,
         "headers": [
+            {
+                "source": "/*/.well-known/openid-configuration",
```
travjenkins, line 4530: Just making sure.... this source should start with `/` ?

I would think this would be like below `**/*/.well-known/openid-configuration`

PR #914, opened 2025-07-28
Make some strapi seed data adaptations

Brenosalv:
Related strapi admin PR: https://github.com/estuary/strapi-admin/pull/94

## Changes

-   Use estuary vendor `slugKey` instead of `id` to check if it's estuary vendor or not;
-   Off topic: Leverage this PR to fix a text from product page.

## Tests / Screenshots

Before:
image

Then:
image

For the comparison pages, I tested with local Strapi seed data and prod data and they look good:

image

image

### DIFF
--- a/gatsby-node.ts
+++ b/gatsby-node.ts
@@ -143,6 +143,10 @@ const createVendorCompare: CreateHelper = async (
 
     // Loop and process what we got (ignore blogs cause that is big and handled down below)
     if (vendors) {
+        const estuary = vendors.find((vendor) =>
+            vendor.slugKey?.includes('estuary')
+        );
+
         vendors.forEach((xVendor, i) => {
             vendors.slice(i + 1).forEach((yVendor) => {
                 if (xVendor.slugKey && yVendor.slugKey) {
@@ -155,8 +159,7 @@ const createVendorCompare: CreateHelper = async (
                         context: {
                             xVendorId: xVendor.id,
                             yVendorId: yVendor.id,
-                            estuaryVendorId:
-                                'd829928c-c473-5421-ac0a-f03c45b14993',
+                            estuaryVendorId: estuary?.id,
                         },
                     });
                 }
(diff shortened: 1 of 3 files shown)


gatsby-node.ts
```diff
@@ -155,8 +159,7 @@ const createVendorCompare: CreateHelper = async (
                         context: {
                             xVendorId: xVendor.id,
                             yVendorId: yVendor.id,
-                            estuaryVendorId:
-                                'd829928c-c473-5421-ac0a-f03c45b14993',
+                            estuaryVendorId: estuary?.id,
```
Brenosalv, line 162: As this `id` comes only from Strapi in prod, we need to switch to `estuary?.id`.

gatsby-node.ts
```diff
@@ -143,6 +143,10 @@ const createVendorCompare: CreateHelper = async (
 
     // Loop and process what we got (ignore blogs cause that is big and handled down below)
     if (vendors) {
+        const estuary = vendors.find((vendor) =>
+            vendor.slugKey?.includes('estuary')
+        );
+
```
Brenosalv, line 149: Checking the `id` instead of the `slugKey` would be safer, but we need this change because now we use strapi with seed data and I think we can't reproduce the same `id` on both prod and local strapi data. As the `slugKey` can be the same for both, it works.

src/pages/etl-tools/index.tsx
```diff
@@ -62,9 +62,7 @@ export default EtlTools;
 
 export const pageQuery = graphql`
     query GetAllComparisons {
-        estuaryVendor: strapiComparison(
-            id: { eq: "d829928c-c473-5421-ac0a-f03c45b14993" }
-        ) {
+        estuaryVendor: strapiComparison(slugKey: { in: "estuary" }) {
```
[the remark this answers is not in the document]
travjenkins: I think this is fine and in _some_ ways better. Like if we ever add more than `Estuary Flow` as a product and have `Estuary Foo` this auto handles that.

